PR #14, opened 2018-05-08
<README/build_deps: Add pkg-config to required dependencies for build>

bobsaintcool:
I believe that pkg-config is also required to build mako.


emersion:
Thanks!

PR #16, opened 2018-05-08
Allow margin to be specified per direction

vilhalmer:
This adds CSS-style margin syntax parsing. That is, you can provide up to four values to `--margin`, comma-separated. A single value specifies all directions, two specify vertical/horizontal, three specify top/horizontal/bottom, and four specify top/right/bottom/left. This might be overkill, it's up to you if you want to have all four modes or cut out the third.

Let me know what you think! If you like this, I'd also plan on adding this to `--padding`, and doing an option to change the anchor corner. Also, my C was a bit rusty so there might be a few things that are suboptimal.


config.c
```diff
@@ -25,6 +33,66 @@ void finish_config(struct mako_config *config) {
...
+	out->right = values[1];
+	out->bottom = values[2];
+	out->left = values[3];
+	break;
+	}
+	}
+
+	free(components);
+	return error;
+}
```
SirCmpwn: You need to add some indents to this function my friend

vilhalmer: Agh, my spaces to tabs function failed me horribly!

vilhalmer:
Alright, got everything cleaned up. Will work on the man page tomorrow!

config.c
```diff
@@ -26,6 +35,66 @@ void finish_config(struct mako_config *config) {
...
+ * struct will be initialized following the same rules as the CSS "margin"
+ * property. Returns 0 on success, -1 on parsing failure. */
+int parse_directional(const char *directional_string,
+		struct directional *out) {
+	int error = 0;
+	char *components = strdup(directional_string);
+
+	char *token, *saveptr, *endptr = NULL;
+	int32_t values[] = {0, 0, 0, 0};
+	int count = 0;
```
emersion: Since this is a counter, it would make sense to make this a `size_t`. Also please define variables when there are used (here in the `for` loop).

config.c
```diff
@@ -85,7 +154,10 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		} else if (strcmp(name, "height") == 0) {
 			config->height = strtol(optarg, NULL, 10);
 		} else if (strcmp(name, "margin") == 0) {
-			config->margin = strtol(optarg, NULL, 10);
+			if (parse_directional(optarg, &config->margin)) {
+				printf("Unable to parse margins\n\n");
```
emersion: Nitpick: only one `\n`, and use `fprintf(stderr, …);` to print ot the error output

include/config.h
```diff
@@ -4,16 +4,25 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+struct directional {
```
emersion: We generally prefix struct names with `mako_`

vilhalmer:
Took a first stab at updating the man page. I broke the explanation of the directional syntax into a section in anticipation of using it elsewhere.

config.c
vilhalmer: I can't move count into the for loop as it's also used down below. I did move `number` and `endptr` into the loop though.

config.c
vilhalmer: Done!

render.c
```diff
@@ -28,6 +28,11 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
 	int border_size = 2 * config->border_size;
 	int padding_size = 2 * config->padding;
 
+    int inner_margin = config->margin.top;
+    if (config->margin.bottom > config->margin.top) {
+        inner_margin = config->margin.bottom;
+    }
```
emersion: Please use tabs instead of spaces

vilhalmer: I installed vim-sleuth so this will hopefully stop happening. :)

SirCmpwn: I bet @emersion would merge an editorconfig if you wrote one

vilhalmer: I was considering that, so if you're both onboard I will definitely do so!

vilhalmer:
The Github docs [request](https://help.github.com/articles/about-pull-requests/) that you don't force push to a PR branch, but I found this which says that the owner can squash if it's enabled on the repo: https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits

SirCmpwn:
Those docs are nonsense, I force push to PRs basically every day.

render.c
emersion: Yeah, +1 to `.editorconfig`

PR #21, opened 2018-05-11
Add config file

### LINKED ISSUE #7: Add a config file for advanced configuration
Should it be dunst-compatible?

SirCmpwn:
Closes #7

### DIFF
--- a/config.c
+++ b/config.c
@@ -1,9 +1,12 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
 #include <getopt.h>
-#include <stdio.h>
+#include <libgen.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <string.h>
+#include <wordexp.h>
+#include <unistd.h>
 
 #include "config.h"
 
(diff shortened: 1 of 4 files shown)


config.c
```diff
@@ -45,6 +49,150 @@ static uint32_t parse_color(const char *color) {
...
+			strcat(config_home, "/.config");
+			setenv("XDG_CONFIG_HOME", config_home, 1);
+			free(config_home);
+		}
+	}
+
+	wordexp_t p;
+	char *path;
+
+	for (int i = 0; i < (int)(sizeof(config_paths) / sizeof(char *)); ++i) {
```
ascent12: Use `size_t i = 0` instead of casting to int.

config.c
```diff
@@ -45,6 +49,150 @@ static uint32_t parse_color(const char *color) {
...
+		free(path);
+		return 0;
+	}
+	const char *base = basename(path);
+
+	int lineno = 0;
+	while (!feof(f)) {
+		char *line = NULL;
+		size_t n = 0;
+		ssize_t l = getline(&line, &n, f);
```
ascent12: `getline` can reuse the allocation, so you could get away with a lot few allocations by moving the variables outside of the loop.
```c
size_t l;
char *line = NULL;
size_t n = 0;
while ((l = getline(&line, &n, f)) > 0) {
    // Do stuff
}
free(line);
```

config.c
```diff
@@ -45,6 +49,150 @@ static uint32_t parse_color(const char *color) {
...
+int load_config_file(struct mako_config *config) {
+	char *path = get_config_path();
+	if (!path) {
+		return 0;
+	}
+
+	FILE *f = fopen(path, "r");
+	if (!f) {
+		free(path);
+		return 0;
```
emersion: `get_config_path` checks that the file exists, so any error opening the file should be fatal?

config.c
```diff
@@ -45,6 +49,150 @@ static uint32_t parse_color(const char *color) {
...
+			path = strdup(p.we_wordv[0]);
+			wordfree(&p);
+			if (file_exists(path)) {
+				return path;
+			}
+			free(path);
+		}
+	}
+
+	return NULL; // Not reached
```
emersion: This comment is misleading: if neither paths exist, the statement is reached.

emersion:
Ah, and follow-up issue: add `makoctl reload`

Issue #3, opened 2018-05-06
Style differently urgent/low priority notifications


emersion:
Maybe implement some sort of filtering system, e..g to be able to style differently notifications coming from a specific app?

Depends on #7, I don't want to add CLI flags for this.

vilhalmer:
I would propose implementing this as a combination of mpv's concept of profiles and sway's criteria syntax. So, for example, in this configuration Firefox notifications inherit the global margin and padding, but are orange. Anything with high urgency that also has at least one action defined has red text:
```
margin=30
padding=10
background-color=#eeeeee

[app-name="Firefox"]
background-color=#dd9b59

[urgency=high actionable]
text-color=#ff00000
```

This will require pcre support if we want to have matching on the summary and body, but that could wait until a future PR.

emersion:
+1 to this design, looks great

PR #26, opened 2018-05-16
add hidden notification count #25

adrubesh:
I went ahead and attempted this. It just simply adds the hidden notification count to the last shown notification in the list. Let me know if there are any issues. 

(I'm Symb0l in the sway-devel irc btw)

### DIFF
--- a/include/render.h
+++ b/include/render.h
@@ -3,18 +3,6 @@
 
 struct mako_state;
 
-enum mako_render_text_alignment {
-	MAKO_RENDER_TEXT_ALIGN_RIGHT = 1,
-	MAKO_RENDER_TEXT_ALIGN_LEFT = 2,
-	MAKO_RENDER_TEXT_ALIGN_CENTER = 3 
-};
-
-void cairo_clear(cairo_t *cairo, int x, int y, int width, int height);
-PangoLayout *get_large_text_layout(cairo_t *cairo, struct mako_config *config, int max_width, int max_height, char *text);
-PangoLayout *get_simple_text_layout(cairo_t *cairo, char *font, int max_width, int max_height);
-void pango_render_layout(cairo_t *cairo, PangoLayout *layout, struct mako_config *config, int offset, uint32_t alignment);
-void set_pango_font(PangoLayout *layout, char *font);
-void set_pango_text(PangoLayout *layout, bool markup, char *text);
 int render(struct mako_state *state, struct pool_buffer *buffer);
 
 #endif
--- a/include/render.h
+++ b/include/render.h
@@ -3,6 +3,18 @@
 
 struct mako_state;
 
+enum mako_render_text_alignment {
+	MAKO_RENDER_TEXT_ALIGN_RIGHT = 1,
+	MAKO_RENDER_TEXT_ALIGN_LEFT = 2,
+	MAKO_RENDER_TEXT_ALIGN_CENTER = 3 
+};
+
+void cairo_clear(cairo_t *cairo, int x, int y, int width, int height);
+PangoLayout *get_large_text_layout(cairo_t *cairo, struct mako_config *config, int max_width, int max_height, char *text);
+PangoLayout *get_simple_text_layout(cairo_t *cairo, char *font, int max_width, int max_height);
+void pango_render_layout(cairo_t *cairo, PangoLayout *layout, struct mako_config *config, int offset, uint32_t alignment);
+void set_pango_font(PangoLayout *layout, char *font);
+void set_pango_text(PangoLayout *layout, bool markup, char *text);
 int render(struct mako_state *state, struct pool_buffer *buffer);
 
 #endif
--- a/render.c
+++ b/render.c
@@ -114,9 +114,46 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
 		notif->hotspot.width = notif_width;
 		notif->hotspot.height = notif_height;
 
-		g_object_unref(layout);
-
 		++i;
+		if(i == (size_t)config->max_visible &&
+				i < (size_t)wl_list_length(&state->notifications) ) {
+
+			int hidden = wl_list_length(&state->notifications) - i;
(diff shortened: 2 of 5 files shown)


emersion:
Thanks for you PR!

Is it possible to use the same code to draw a notification and to draw the notification count?

render.c
```diff
@@ -114,9 +114,46 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
 		notif->hotspot.width = notif_width;
 		notif->hotspot.height = notif_height;
 
-		g_object_unref(layout);
-
 		++i;
+		if(i == (size_t)config->max_visible &&
+				i < (size_t)wl_list_length(&state->notifications) ) {
```
SirCmpwn: Style issue, add space before the ( and remove space before the )

adrubesh:
Ahh okay ignore that. Overlooked that option.

adrubesh:
So, I did quite a bit of refactoring. Probably too much for this issue alone but I tried to clean up some of the rendering parts and make some pieces more flexible.

include/render.h
```diff
@@ -3,6 +3,18 @@
 
 struct mako_state;
 
+enum mako_render_text_alignment {
```
emersion: No need for this, pango already has it: https://developer.gnome.org/pango/stable/pango-Layout-Objects.html#pango-layout-set-alignment

emersion:
Yeah, I'm not sure I like this. This code will only ever render notifications (ie. boxes with a border and some text inside), so I don't see the value of abstracting things away. Also when we'll add HiDPI support the rendering code will become more complicated to understand, so I really want to keep it as simple as possible.

When I was talking about adding a function that renders notifications, I was thinking of something among the lines of:

```c
int render_notification(struct mako_state *state, int y, const char *text, bool markup);
```

Which would return the height of the renderer box (`notif_height`).

This function can be reused to render the last fake notification at the bottom with the number of hidden notifications.

Don't worry about styling it differently, this will be handled in #3.

render.c
```diff
@@ -97,26 +187,42 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
...
 		// Update hotspot
 		notif->hotspot.x = 0;
 		notif->hotspot.y = notif_y;
 		notif->hotspot.width = notif_width;
 		notif->hotspot.height = notif_height;
 
-		g_object_unref(layout);
-
 		++i;
+		if (i == (size_t)config->max_visible &&
```
emersion: Can you move this code out of the loop?

adrubesh:
Sorry about that just my coding habbits and I was conflicted with the styling aspect as well. Since that is not an issue, I believe my latest change is more inline with what you wanted.

render.c
```diff
@@ -5,14 +5,83 @@
 #include "notification.h"
 #include "render.h"
 
-static void set_cairo_source_u32(cairo_t *cairo, uint32_t color) {
+
```
emersion: Style nitpick: can you remove these two blank lines?

render.c
```diff
@@ -5,14 +5,83 @@
...
+
+void set_cairo_source_u32(cairo_t *cairo, uint32_t color) {
```
emersion: This function should be `static` because it's only used in this file.

render.c
```diff
@@ -5,14 +5,83 @@
...
 	cairo_set_source_rgba(cairo,
 		(color >> (3*8) & 0xFF) / 255.0,
 		(color >> (2*8) & 0xFF) / 255.0,
 		(color >> (1*8) & 0xFF) / 255.0,
 		(color >> (0*8) & 0xFF) / 255.0);
 }
 
+int render_notification(cairo_t *cairo, struct mako_state *state, const char *text, int offset_y) {
```
emersion: This on should also be `static`

render.c
```diff
@@ -39,89 +105,45 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
...
-
-		// Render text
-		set_cairo_source_u32(cairo, config->colors.text);
-		cairo_move_to(cairo, config->border_size + config->padding,
-			notif_y + config->border_size + config->padding);
-		pango_cairo_update_layout(cairo, layout);
-		pango_cairo_show_layout(cairo, layout);
+
+		int notif_height = render_notification(cairo, state, text, notif_y);
+		height = notif_y +  notif_height;
```
emersion: Style nitpick: there are two spaces after `+`

render.c
```diff
@@ -39,89 +105,45 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
...
 			break;
 		}
 	}
 
+	if (wl_list_length(&state->notifications) > config->max_visible) {
+		int hidden = wl_list_length(&state->notifications) - config->max_visible;
+		int hidden_ln = snprintf(NULL, 0, "[%d]", hidden);
+
+		char *hidden_text;
+		hidden_text = malloc(hidden_ln + 1);
```
emersion: This is never free'd. Instead you can use VLAs:

```c
char hidden_text[hidden_ln + 1];
```

render.c
```diff
@@ -39,89 +105,45 @@ int render(struct mako_state *state, struct pool_buffer *buffer) {
...
+		snprintf(hidden_text, hidden_ln + 1, "[%d]", hidden);
+
+		int hidden_height = render_notification(cairo, state, hidden_text, height);
```
emersion: We need to add `inner_margin` to `height` before this line, otherwise the hidden count notification collides with the previous one.

adrubesh:
You have a much better eye for the style issues than I do. Thanks for pointing them out.

render.c
adrubesh: The reason for this was that personally I liked the look of the count being closer to the notifications than the spacing provided by `inner_margin` but I added your change.

emersion: Ah, I see. In the future, this can probably be added back as a config option somehow.

emersion:
Opened a follow-up issue: #29

PR #30, opened 2018-05-17
Create a new cairo path for each notification

vilhalmer:
I've still been investigating why borders get thicker in between adjacent notifications. In the process, I found that they were accidentally being drawn multiple times along each notification. This was invisible since the fill is then drawn over top. I still haven't solved the original issue, but I figured this was worth getting fixed in the meantime.

I don't know enough cairo to know why this is the correct solution instead of cairo_close_path, but this is the one that worked.

### DIFF
--- a/render.c
+++ b/render.c
@@ -102,6 +102,11 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	pango_cairo_update_layout(cairo, layout);
 	pango_cairo_show_layout(cairo, layout);
 
+	// Reset the cairo path between each notification, otherwise subsequent
+	// borders will be drawn along all notifications, causing them to become
+	// visually thicker.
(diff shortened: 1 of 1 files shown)


emersion:
Hmm, it seems like `cairo_close_path` would be the correct solution, because `cairo_move_to` is documented to create a new path: https://www.cairographics.org/manual/cairo-Paths.html#cairo-move-to

Can you try if it fixes the issue?

vilhalmer:
It doesn't, though I couldn't tell you why. If you want to leave this sitting here for the moment though, there's another half of this issue that I'm still working on involving the surface not being cleared properly once it reaches its maximum height. Maybe they're related, and this will go away on its own after I figure that one out.

vilhalmer:
Fixed for real by bf49b784, closing.

PR #34, opened 2018-05-20
[WIP] Per-notification style

vilhalmer:
Work towards #3. Posting it early to get feedback on the structure I've chosen, and also just because I've got a lot going on in real life and I don't want to just have radio silence over the next couple weeks.

List of things done, and things left to do:
- [x] Render each notification using its associated config
- [x] Add `mako_criteria` and match logic for each applicable notification field
    - [x] `app_name`
    - [x] `app_icon`
    - [x] `actionable` (whether the actions list is nonempty)
    - [x] `urgency`
    - [x] `category`
    - [x] `desktop_entry`
    - ~`summary`~ (pcre)
    - ~`body`~ (pcre)
- [ ] Add criteria support to configuration parser
    - [ ] Convert hidden to a criteria
- [ ] Store criteria structs on `mako_state`
- [ ] Rework global configuration to be an empty criteria
- [ ] Check for criteria matches in `create_notification` and set the notification's config
- [ ] Adjust rendering for differently-sized notifications
    - [ ] Calculate maximum border + padding + margin for size of surface (just resize on the fly?)
    - [ ] Anchor smaller notifications along the same edge as the surface
- [ ] Allow choosing output per-notification
    - [ ] Loop over outputs in send_frame
    - [ ] Make render take the output (replacing `scale`), and only draw appropriate notifications

Things I'm not sure about:
- [ ] Can we match on how close a timer is to expiring? (e.g., have notifications visibly age.)
- [ ] Should `max_visible` be settable in criteria? We would have to decide on how the criteria setting interacts with the global one.
- [ ] How should notifications that match multiple criteria be handled? Does the last defined criteria "win" if multiple set the same option? Or do we want some other priority?

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -1,0 +1,33 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <wayland-client.h>
+#include "mako.h"
+#include "criteria.h"
+#include "notification.h"
+
+struct mako_criteria *create_criteria(struct mako_state *state) {
+	struct mako_criteria *criteria = calloc(1, sizeof(struct mako_criteria));
+	if (criteria == NULL) {
+		fprintf(stderr, "allocation failed\n");
+		return NULL;
+	}
+
+	// Copy the global configuration, which will then be overridden as needed
+	// during the configuration of this criteria.
+	*criteria->config = state->config;
+	return criteria;
+}
+
+void finish_criteria(struct mako_criteria *criteria) {
+	free(criteria->app_name);
+	free(criteria->app_icon);
+	free(criteria->category);
+	free(criteria->desktop_entry);
+}
+
+bool match_criteria(struct mako_criteria *criteria,
+		struct mako_notification *notif) {
+	return false;
+}
--- a/criteria.c
+++ b/criteria.c
@@ -29,5 +29,37 @@ void finish_criteria(struct mako_criteria *criteria) {
 
 bool match_criteria(struct mako_criteria *criteria,
 		struct mako_notification *notif) {
-	return false;
+	struct mako_criteria_spec spec = criteria->specified;
+
+	if (spec.app_name &&
+			strcmp(criteria->app_name, notif->app_name) != 0) {
+		return false;
+	}
+
+	if (spec.app_icon &&
+			strcmp(criteria->app_icon, notif->app_icon) != 0) {
+		return false;
+	}
+
+	if (spec.actionable &&
+			criteria->actionable == wl_list_empty(&notif->actions)) {
+		return false;
+	}
+
+	if (spec.urgency &&
+			criteria->urgency != notif->urgency) {
+		return false;
+	}
+
+	if (spec.category &&
+			strcmp(criteria->category, notif->category) != 0) {
+		return false;
+	}
+
+	if (spec.desktop_entry &&
+			strcmp(criteria->desktop_entry, notif->desktop_entry) != 0) {
+		return false;
+	}
+
+	return true;
 }
--- a/include/config.h
+++ b/include/config.h
@@ -24,6 +24,8 @@ enum mako_sort_criteria {
 };
 
 struct mako_config {
+	// Configurable per notification
+
 	char *font;
 	int32_t width, height;
 	int32_t padding;
@@ -32,11 +34,6 @@ struct mako_config {
 	char *format;
 	bool actions;
 	struct mako_directional margin;
-	int32_t max_visible;
-	char *output;
-	char *hidden_format;
-	uint32_t sort_criteria; //enum mako_sort_criteria
-	uint32_t sort_asc;
 
 	int default_timeout; // in ms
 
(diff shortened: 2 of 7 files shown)


include/config.h
```diff
@@ -19,6 +19,8 @@ enum mako_button_binding {
 };
 
 struct mako_config {
+	// Configurable per notification
```
emersion: Maybe we should add a `struct mako_notification_config` and add a `struct mako_notification_config default` field to `mako_config`

include/criteria.h
```diff
@@ -0,0 +1,45 @@
...
+	bool app_icon;
+	bool actionable;
+	bool urgency;
+	bool category;
+	bool desktop_entry;
+};
+
+struct mako_criteria {
+	// Configuration to apply to matches:
+	struct mako_config *config;
```
emersion: This can be an embedded struct instead (`struct mako_config config`).

include/criteria.h
```diff
@@ -0,0 +1,45 @@
...
+#include <stdint.h>
+#include "notification.h"
+
+struct mako_config;
+
+// Stores whether or not each field was part of the criteria specification, so
+// that, for example, "not actionable" can be distinguished from "don't care".
+// This is unnecessary for string fields, but it's best to just keep it
+// consistent.
+struct mako_criteria_spec {
```
emersion: Instead, we could have an `enum mako_criteria_bool` with three variants: none, true, false.

render.c
```diff
@@ -86,15 +84,15 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	// Render border
 	set_source_u32(cairo, config->colors.border);
 	cairo_set_line_width(cairo, border_size * scale);
-	set_rectangle(cairo, 0, offset_y, state->width, notif_height, scale);
+	set_rectangle(cairo, 0, offset_y, config->width, notif_height, scale);
```
emersion: Hmm, this still needs to be `state`, because even if the compositor gave us less width than we asked, we still want to render ourselves correctly.

render.c
```diff
@@ -86,15 +84,15 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
...
 	cairo_stroke(cairo);
 
 	// Render background
 	set_source_u32(cairo, config->colors.background);
 	cairo_set_line_width(cairo, 0);
 	set_rectangle(cairo,
 		config->border_size, offset_y + config->border_size,
-		state->width - border_size, notif_height - border_size, scale);
+		config->width - border_size, notif_height - border_size, scale);
```
emersion: Ditto

criteria.c
```diff
@@ -0,0 +1,65 @@
...
+		return NULL;
+	}
+
+	// Copy the global configuration, which will then be overridden as needed
+	// during the configuration of this criteria.
+	*criteria->config = state->config;
+	return criteria;
+}
+
+void finish_criteria(struct mako_criteria *criteria) {
```
emersion: If you have a `create_criteria` function, you need `destroy_criteria` instead of `finish_criteria`. The pattern is:

* `create` and `destroy` allocates and de-allocates the structs and their fields
* `init` and `finish` only allocates and de-allocates the fields inside the structs, not the structs themselves

include/config.h
vilhalmer: I've been debating that also. It really depends on how many things we expect to accumulate over time that don't make sense to override per-notification. (Hence the item about max_visible in the "things I'm not sure about" list.)

include/criteria.h
vilhalmer: I had that originally, but then immediately ran into needing something for `urgency` as well.

emersion:
PCRE stuff can be added in another PR if you want to keep this one small

> Store criteria structs somewhere

Yeah, `mako_state` is a good place. You need to add a `wl_list` for `mako_criteria` for this.

>Check for criteria matches in handle_notify

Could do that in `create_notification`

>Adjust rendering for differently-sized notifications

This will be tricky. Can be left as a TODO for a future PR if you want.

render.c
vilhalmer: OK, I wasn't sure about that. I'll pass both.

include/criteria.h
emersion: Hmm, I see. We could then store bools alongside the other fields, but this is fine too.

vilhalmer: I liked having the symmetry in `match_criteria`, but if it becomes too annoying to maintain these duplicate definitions, it should be easy enough to flatten.

emersion:
Hey, any news about this? :)

vilhalmer:
I'm in the process of moving so I haven't had time to work on it, but I will be back to normal life in a couple days and back to full speed!

emersion:
Thanks, no worries!

vilhalmer:
Just rewriting some history here to get all the fancy new toys. :)

The new `[hidden]` section should translate nicely into a criteria (with just a little bit of handwaving), which means it should be easy to style it however you want as well.

Regarding whether or not to split notification config into a new struct separate from `mako_config`, I think we've now accumulated enough truly global options that it makes sense. I don't see how the priority sorting that just went in could be useful to configure with criteria, unless we wanted to get really complex and support sorting groups of notifications individually and then sorting the groups based on the global setting. That immediately gets into cases like what it means to sort a group by priority, etc, which don't seem like a lot of fun to handle.

vilhalmer:
Closing in favor of #43.

PR #35, opened 2018-05-21
Hidden notification format

adrubesh:
I took your idea of the generic format functions and implemented `format_notif_text` and `format_state_text`. This would allow us to format strings based on state data and notification data. Next, I am going to work on adding something to the config so you can specify hidden notification format.

### DIFF
--- a/config.c
+++ b/config.c
@@ -41,6 +41,7 @@ void init_config(struct mako_config *config) {
 void finish_config(struct mako_config *config) {
 	free(config->font);
 	free(config->format);
+	free(config->hidden_format);
 	free(config->output);
 }
 
--- a/config.c
+++ b/config.c
@@ -19,7 +19,7 @@ void init_config(struct mako_config *config) {
 	config->border_size = 1;
 	config->markup = true;
 	config->format = strdup("<b>%s</b>\n%b");
-	config->hidden_format = strdup("[%h]");
+	config->hidden_format = strdup("%t[%h]");
 	config->actions = true;
 
 	config->margin.top = 10;
--- a/config.c
+++ b/config.c
@@ -19,12 +19,14 @@ void init_config(struct mako_config *config) {
 	config->border_size = 1;
 	config->markup = true;
 	config->format = strdup("<b>%s</b>\n%b");
+	config->hidden_format = strdup("[%h]");
 	config->actions = true;
 
 	config->margin.top = 10;
 	config->margin.right = 10;
 	config->margin.bottom = 10;
 	config->margin.left = 10;
+	config->hidden_margin = 3;
 
 	config->max_visible = 5;
 	config->output = strdup("");
--- a/config.c
+++ b/config.c
@@ -161,6 +161,10 @@ static int apply_config_option(struct mako_config *config,
 		free(config->format);
 		config->format = strdup(value);
 		return 0;
+	} else if (strcmp(name, "hidden-format") == 0) {
+		free(config->hidden_format);
+		config->hidden_format = strdup(value);
+		return 0;
 	} else if (strcmp(name, "max-visible") == 0) {
 		config->max_visible = strtol(value, NULL, 10);
 		return 0;
@@ -282,6 +286,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"border-color", required_argument, 0, 0},
 		{"markup", required_argument, 0, 0},
 		{"format", required_argument, 0, 0},
+		{"hidden-format", required_argument, 0, 0},
 		{"max-visible", required_argument, 0, 0},
 		{"default-timeout", required_argument, 0, 0},
 		{"output", required_argument, 0, 0},
--- a/config.c
+++ b/config.c
@@ -26,7 +26,6 @@ void init_config(struct mako_config *config) {
 	config->margin.right = 10;
 	config->margin.bottom = 10;
 	config->margin.left = 10;
-	config->hidden_margin = 3;
 
 	config->max_visible = 5;
 	config->output = strdup("");
--- a/notification.c
+++ b/notification.c
@@ -329,10 +329,10 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
 	case MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION:;
 		struct mako_action *action;
 		wl_list_for_each(action, &notif->actions, link) {
-		       if (strcmp(action->key, DEFAULT_ACTION_KEY) == 0) {
+			if (strcmp(action->key, DEFAULT_ACTION_KEY) == 0) {
 			       notify_action_invoked(action);
 			       break;
-		       }
+			}
 		}
 		close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
 		break;
--- a/notification.c
+++ b/notification.c
@@ -174,7 +174,7 @@ const char* format_notif_text(char variable, void *data) {
 	return value;
 }
 
-size_t format_text(const char *format, char *buf, mako_format_func_t func, void *data) {
+size_t format_text(const char *format, char *buf, mako_format_func_t format_func, void *data) {
 	size_t len = 0;
 
 	const char *last = format;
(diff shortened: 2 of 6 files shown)


config.c
```diff
@@ -19,12 +19,14 @@ void init_config(struct mako_config *config) {
...
 	config->markup = true;
 	config->format = strdup("<b>%s</b>\n%b");
+	config->hidden_format = strdup("[%h]");
 	config->actions = true;
 
 	config->margin.top = 10;
 	config->margin.right = 10;
 	config->margin.bottom = 10;
 	config->margin.left = 10;
+	config->hidden_margin = 3;
```
emersion: Hum, is it possible to do not add this? I'd prefer to do it via #3, with a config like:

```ini
[hidden-count]
margin=3
```

notification.c
```diff
@@ -135,8 +135,104 @@ static size_t escape_markup(const char *s, char *buf) {
 	return len;
 }
 
-size_t format_notification(struct mako_notification *notif, const char *format,
-		char *buf) {
+static char *notification_hidden_count(struct mako_state *state) {
+	int hidden = (wl_list_length(&state->notifications) -
+			state->config.max_visible);
+
+	int hidden_ln = snprintf(NULL, 0, "%d",  hidden);
```
emersion: `hidden_ln` should be a `size_t`

notification.c
```diff
@@ -135,8 +135,104 @@ static size_t escape_markup(const char *s, char *buf) {
...
+
+	return hidden_text;
+}
+
+const char *format_state_text(char variable, void *data) {
+	struct mako_state *state = (struct mako_state *)data;
+	const char *value = NULL;
+	switch (variable) {
+	case 'h':
+		value = notification_hidden_count(state);
```
emersion: Can simplify these `switch` statements with:

```c
switch (variable) {
case 'h:
    return notification_hidden_count(state);
}
```

render.c
```diff
@@ -150,6 +150,8 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 
 		int notif_height =
 			render_notification(cairo, state, text, notif_y, scale);
+		free(text);
```
emersion: Good catch

notification.c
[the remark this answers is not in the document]
adrubesh: What if I did something like:
```
const char* format_notif_text(char variable, void *data) {
	struct mako_notification *notif = (struct mako_notification *)data;
	char *value = NULL;
	switch (variable) {
	case 'a':
		value = strdup(notif->app_name);
		break;
	}
	return value;
}

const char *format_state_text(char variable, void *data) {
	struct mako_state *state = (struct mako_state *)data;
	const char *value = NULL;
	switch (variable) {
	case 'h':
		value = notification_hidden_count(state);
		break;
	}
	return value;
}
```

Then `free(value)` in `format_text` after the `memcpy`?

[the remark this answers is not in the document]
emersion: Yeah, that would work, but that would require an extra allocation. Probably better to do it this way, you're right.

Note that the function signature should become `char *format_state_text(char variable, void *data` (`const char *` is for something you don't need to free).

notification.c
```diff
@@ -135,8 +137,109 @@ static size_t escape_markup(const char *s, char *buf) {
...
+		size_t hidden_ln = snprintf(NULL, 0, "%d", hidden);
+		value = malloc(hidden_ln + 1);
+		if(!value) {
+			break;
+		}
+		snprintf(value, hidden_ln + 1, "%d", hidden);
+		return value;
+	case 't':;
+		int count = wl_list_length(&state->notifications);
+		size_t total_ln = snprintf(NULL, 0, "%d", count);
```
adrubesh: Was thinking about making this "int to char array" into a function since I've already used it twice.

render.c
```diff
@@ -169,7 +169,7 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 
 	if (wl_list_length(&state->notifications) > config->max_visible) {
 
-		height += config->hidden_margin;
+		height += 3;
```
emersion: Hmm, why?

adrubesh: No reason, reverted it back to inner_margin for now.

config.c
```diff
@@ -19,6 +19,7 @@ void init_config(struct mako_config *config) {
 	config->border_size = 1;
 	config->markup = true;
 	config->format = strdup("<b>%s</b>\n%b");
+	config->hidden_format = strdup("%t[%h]");
```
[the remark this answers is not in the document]
emersion: Yeah, since we don't yet have #3, I think it's okay. This will probably be changed in the future though.

adrubesh:
@emersion Done with this. Let me know what you think.

notification.c
emersion: +1 for implementing `asprintf`

This function is not part of the standard C library - it's a GNU extension. We don't use GNU extensions because they're not portable and not part of the standard.

notification.c
```diff
@@ -135,8 +137,50 @@ static size_t escape_markup(const char *s, char *buf) {
...
+	}
+	return value;
+}
+
+char *format_notif_text(char variable, void *data) {
+	struct mako_notification *notif = data;
+	char *value = NULL;
+	switch (variable) {
+	case 'a':
+		value = strdup(notif->app_name);
```
emersion: You can just `return strdup(notif->app_name)` instead of setting `value`.

notification.c
```diff
@@ -135,8 +137,50 @@ static size_t escape_markup(const char *s, char *buf) {
...
+		int count = wl_list_length(&state->notifications);
+		size_t total_ln = snprintf(NULL, 0, "%d", count);
+		value = malloc(total_ln + 1);
+		if(!value) {
+			break;
+		}
+		snprintf(value, total_ln + 1, "%d", count);
+		return value;
+	}
+	return value;
```
emersion: No need fro this `value` variable - just `return NULL`

notification.c
```diff
@@ -157,22 +201,13 @@ size_t format_notification(struct mako_notification *notif, const char *format,
...
-			break;
-		case 'b':
-			value = notif->body;
-			markup = true;
-			break;
+		} else {
+			value =	format_func(current[1], data);
 		}
 		if (value == NULL) {
 			value = "";
```
emersion: Should be `value = strdup("")` since we `free` it later

notification.c
adrubesh: Should this function be added to notification.c? Or would it be better to create a tool/utility header so so we include elsewhere if needed?

emersion: Let's just add it here as a static function for now, and export it if we need it elsewhere.

notification.c
```diff
@@ -157,25 +209,20 @@ size_t format_notification(struct mako_notification *notif, const char *format,
...
-			break;
+		} else {
+			value =	format_func(current[1], data);
 		}
 		if (value == NULL) {
-			value = "";
+			value = strdup("");
+		}
+
+		if (current[1] == 'b') {
```
adrubesh: I'm not sure how to handle setting `markup` yet. You currently had it just for handling the body, so I just included it here like this for now, but it should probably be handled differently.

notification.c
```diff
@@ -137,47 +137,55 @@ static size_t escape_markup(const char *s, char *buf) {
 	return len;
 }
 
+static char *mako_asprintf(const char *fmt, ...)
+{
```
SirCmpwn: Style: put this { on the previous line

notification.c
emersion: Basically we want to interpret markup (ie. tags like `<b>`) only in the body of the notification. If the notification body is `<b>Hi there</b>`, we want to print "Hi there" in bold. If the notification summary is `<b>Hi there</b>`, we want to print "<b>Hi there</b>" as is.

Maybe `&markup` can be passed as an argument to `format_func`.

notification.c
```diff
@@ -157,25 +209,16 @@ size_t format_notification(struct mako_notification *notif, const char *format,
...
 		len += chunk_len;
 
-		const char *value = NULL;
+		char *value = NULL;
 		bool markup = false;
-		switch (current[1]) {
-		case '%':
+
+		if (current[1] == '%') { 
 			value = "%";
```
emersion: Ah, need to `strdup` this one too

emersion:
There's currently no way to setting `config->hidden_format`. A command-line argument and a config option should be added.

adrubesh:
I was unsure if I needed to or not, since you mentioned we'll probably change it later on. But I'll go ahead an add for now then.

emersion:
Nice work, thanks!

Issue #24, opened 2018-05-14
Add an option to sort notifications by priority


emersion:

adrubesh:
How would you like to sort multile notifications of the same priority? Would it be viable to add a `created_at` timespec value to the notification struct so that multiple, high priority notificaitons will be sorted by time created?

emersion:
Fixed in #40

PR #40, opened 2018-05-28
Added priority sort options 

adrubesh:
Test it out with: 
```
./build/mako --max-visible 11 --sort +time --sort -priority
```

then to test, you can run the following and see how they get inserted: 
```
 for i in {1..10}; do  \
  URGENCY="";         \
  if [ $((i % 3)) -eq 0 ]; then \
    URGENCY="CRITICAL"; \
  elif [ $((i % 2)) -eq 0 ]; then \
    URGENCY="NORMAL";  \
  else \
    URGENCY="LOW"; \
  fi; \
  eval "notify-send -u $URGENCY -a \"test\" \"$URGENCY - $i\""; \
  sleep 0.5; \
done
```

### DIFF
--- a/config.c
+++ b/config.c
@@ -180,16 +180,16 @@ static int apply_config_option(struct mako_config *config,
 	} else if (strcmp(name, "sort") == 0) {
 		if (strcmp(value, "+priority") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_URGENCY;
-			config->sort_asc |= MAKO_SORT_ASC_URGENCY;
+			config->sort_asc |= MAKO_SORT_CRITERIA_URGENCY;
 		} else if (strcmp(value, "-priority") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_URGENCY;
-			config->sort_asc &= ~MAKO_SORT_ASC_URGENCY;
+			config->sort_asc &= ~MAKO_SORT_CRITERIA_URGENCY;
 		} else if (strcmp(value, "+time") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_TIME;
-			config->sort_asc |= MAKO_SORT_ASC_TIME;
+			config->sort_asc |= MAKO_SORT_CRITERIA_TIME;
 		} else if (strcmp(value, "-time") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_TIME;
-			config->sort_asc &= ~MAKO_SORT_ASC_TIME;
+			config->sort_asc &= ~MAKO_SORT_CRITERIA_TIME;
 		}
 		return 0;
 	} else {
@@ -306,6 +306,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"hidden-format", required_argument, 0, 0},
 		{"max-visible", required_argument, 0, 0},
 		{"default-timeout", required_argument, 0, 0},
+		{"output", required_argument, 0, 0},
 		{"sort", required_argument, 0, 0},
 		{0},
 	};
--- a/config.c
+++ b/config.c
@@ -176,17 +176,14 @@ static int apply_config_option(struct mako_config *config,
 		free(config->output);
 		config->output = strdup(value);
 		return 0;
-	} else if (strcmp(name, "urgency") == 0) {
-		if (strcmp(value, "asc") == 0) {
+	} else if (strcmp(name, "sort") == 0) {
+		if (strcmp(value, "+priority") == 0) {
 			config->sort_direction |= (MAKO_SORT_DIRECTION_URGENCY | MAKO_SORT_DIRECTION_URGENCY_ASC);
-		} else if (strcmp(value, "des") == 0) {
+		} else if (strcmp(value, "-priority") == 0) {
 			config->sort_direction |= (MAKO_SORT_DIRECTION_URGENCY);
-		}
-		return 0;
-	} else if (strcmp(name, "time") == 0) {
-		if (strcmp(value, "asc") == 0) {
+		} else if (strcmp(value, "+time") == 0) {
 			config->sort_direction |= (MAKO_SORT_DIRECTION_TIME_ASC);
-		} else if (strcmp(value, "des") == 0) {
+		} else if (strcmp(value, "-time") == 0) {
 			config->sort_direction &= ~(MAKO_SORT_DIRECTION_URGENCY);
 		}
 		return 0;
@@ -304,9 +301,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"hidden-format", required_argument, 0, 0},
 		{"max-visible", required_argument, 0, 0},
 		{"default-timeout", required_argument, 0, 0},
-		{"output", required_argument, 0, 0},
-		{"urgency", required_argument, 0, 0},
-		{"time", required_argument, 0, 0},
+		{"sort", required_argument, 0, 0},
 		{0},
 	};
 
--- a/config.c
+++ b/config.c
@@ -21,7 +21,8 @@ void init_config(struct mako_config *config) {
 	config->format = strdup("<b>%s</b>\n%b");
 	config->hidden_format = strdup("%t[%h]");
 	config->actions = true;
-	config->sort_direction = 0;
+	config->sort_criteria = MAKO_SORT_CRITERIA_TIME;
+	config->sort_asc = 0;
 
 	config->margin.top = 10;
 	config->margin.right = 10;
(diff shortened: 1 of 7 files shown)


config.c
```diff
@@ -175,6 +176,20 @@ static int apply_config_option(struct mako_config *config,
 		free(config->output);
 		config->output = strdup(value);
 		return 0;
+	} else if (strcmp(name, "urgency") == 0) {
```
adrubesh: There's probably a better way to do this. Let me know if you have any ideas.

notification.c
```diff
@@ -286,3 +286,47 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+	}
+
+	return notifications;
+}
+
+void insert_notification(struct mako_state *state, struct mako_notification *notif) {
+	struct mako_config *config = &state->config;
+	struct wl_list *insert_node;
+
+	//This works but I don't like the way it's done. Can probably use some work.
```
adrubesh: Same here. Probably a better way to do this. It works but it's ugly (to me anyway).

config.c
emersion: Hmm, is it possible to have a `sort` option instead, as described in https://github.com/emersion/mako/issues/24#issuecomment-389888975 ? Let me know if you have better ideas :)

include/config.h
```diff
@@ -18,6 +18,12 @@ enum mako_button_binding {
 	MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION,
 };
 
+enum mako_sort_direction {
+	MAKO_SORT_DIRECTION_TIME_ASC = 1,
+	MAKO_SORT_DIRECTION_URGENCY_ASC = 2,
+	MAKO_SORT_DIRECTION_URGENCY = 4,
```
emersion: Hmm, so I guess this is where there are some limitations to using a bitfield for this.

Maybe we could have a `sort` bitfield, which specifies which criteria is sorted, and another `sort_asc` bitfield, which specifies which criteria are sorted in ascending order (the default being descending). Thoughts?

adrubesh:
Overall, is this the functionality you were expecting for this issue?

config.c
adrubesh: I added these options. You can specify `--sort` mutiple times with `+/-time` and `+/-priority`

emersion: Hmm? Correct me if I'm mistaken, but this code doesn't seem to do this. It seems to do things like `--time asc --urgency des` instead.

include/config.h
emersion: >Maybe we could have a sort bitfield, which specifies which criteria is sorted, and another sort_asc bitfield, which specifies which criteria are sorted in ascending order (the default being descending). Thoughts?

emersion:
>Overall, is this the functionality you were expecting for this issue?

Yeah! :)

config.c
adrubesh: Sorry! Forgot to push. :)

config.c
```diff
@@ -289,7 +306,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"hidden-format", required_argument, 0, 0},
 		{"max-visible", required_argument, 0, 0},
 		{"default-timeout", required_argument, 0, 0},
-		{"output", required_argument, 0, 0},
+		{"sort", required_argument, 0, 0},
```
emersion: Hum, can we keep `--output`?

include/config.h
```diff
@@ -18,6 +18,16 @@ enum mako_button_binding {
 	MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION,
 };
 
+enum mako_sort_criteria {
+	MAKO_SORT_CRITERIA_TIME = 1,
+	MAKO_SORT_CRITERIA_URGENCY = 2,
+};
+
+enum mako_sort_asc {
+	MAKO_SORT_ASC_TIME = 1,
```
emersion: No need for two enums here - only one will do

include/config.h
```diff
@@ -30,6 +40,8 @@ struct mako_config {
 	int32_t max_visible;
 	char *output;
 	char *hidden_format;
+	enum mako_sort_criteria sort_criteria;
```
emersion: Since this is a bitfield, this variable can take values that are not in the enum - like `MAKO_SORT_CRITERIA_TIME | MAKO_SORT_CRITERIA_URGENCY`. So this should be a `uint32` instead. Though it's a good practice to add a comment referring to the enum:

```c
uint32_t sort_criteria; // enum mako_sort_criteria
```

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
 		break;
 	}
 }
+
+/*
+ * Searches through the notifications list and returns the next position at which to insert.
+ * If no results for the specified urgency are found, it will return the closest link searching in the direction specifed. (-1 for lower, 1 or upper).
+ */
+static struct wl_list *get_last_notif_by_urgency(struct wl_list *notifications, enum mako_notification_urgency urgency, int direction) {
```
emersion: Style: 80 char limit - please break down this line and prepend two tabs to subsequent lines

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+	enum mako_notification_urgency current = urgency;
+
+	if (wl_list_length(notifications) == 0) {
+		return notifications;
+	}
+
+	while (current <= MAKO_NOTIFICATION_URGENCY_HIGH && current >= MAKO_NOTIFICATION_URGENCY_UNKNOWN) {
```
emersion: Style: 80 char limit

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
 }
+
+/*
+ * Searches through the notifications list and returns the next position at which to insert.
+ * If no results for the specified urgency are found, it will return the closest link searching in the direction specifed. (-1 for lower, 1 or upper).
+ */
+static struct wl_list *get_last_notif_by_urgency(struct wl_list *notifications, enum mako_notification_urgency urgency, int direction) {
+	enum mako_notification_urgency current = urgency;
+
+	if (wl_list_length(notifications) == 0) {
```
emersion: It's better to use `wl_list_empty` instead - it doesn't require to iterate over the whole list

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+	} else if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			(config->sort_asc & MAKO_SORT_ASC_TIME)) {
+			insert_node = state->notifications.prev;
+	} else if (config->sort_criteria & MAKO_SORT_CRITERIA_URGENCY) {
+		int direction = (config->sort_asc & MAKO_SORT_ASC_URGENCY) ? -1 : 1;
+		int offset = 0;
+		if (!(config->sort_asc & MAKO_SORT_ASC_TIME)) {
+			offset = direction;
+		}
+		insert_node = get_last_notif_by_urgency(&state->notifications, notif->urgency + offset, direction);
```
emersion: Style: 80 char limit

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+void insert_notification(struct mako_state *state, struct mako_notification *notif) {
+	struct mako_config *config = &state->config;
+	struct wl_list *insert_node;
+
+	if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			!config->sort_asc) {
+		insert_node = &state->notifications;
+	} else if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			(config->sort_asc & MAKO_SORT_ASC_TIME)) {
+			insert_node = state->notifications.prev;
```
emersion: Style: there's one extra tab here

notification.c
```diff
@@ -286,3 +286,51 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+
+	return notifications;
+}
+
+void insert_notification(struct mako_state *state, struct mako_notification *notif) {
+	struct mako_config *config = &state->config;
+	struct wl_list *insert_node;
+
+	if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			!config->sort_asc) {
```
emersion: `!config->sort_asc` should probably be `!(config->sort_asc & MAKO_SORT_ASC_TIME)`

config.c
adrubesh: Oops. Good catch.

notification.c
adrubesh: wasn't aware of it. thank you!

include/config.h
```diff
@@ -18,6 +18,13 @@ enum mako_button_binding {
 	MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION,
 };
 
+enum mako_sort_criteria {
+	MAKO_SORT_CRITERIA_TIME = 1,
+	MAKO_SORT_CRITERIA_URGENCY = 2,
+	MAKO_SORT_ASC_TIME = 1,
```
emersion: It's pretty weird to have multiple enum items with the same value. I think the last two can be removed.

notification.c
```diff
@@ -286,3 +286,56 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+
+	return notifications;
+}
+
+void insert_notification(struct mako_state *state, struct mako_notification *notif) {
+	struct mako_config *config = &state->config;
+	struct wl_list *insert_node;
+
+	if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+		!(config->sort_asc & MAKO_SORT_ASC_TIME)) {
```
emersion: Can you indent this line with one more tab? The intent is to make it clear what is the condition and what are the instructions to execute:

```c
if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
		!(config->sort_asc & MAKO_SORT_ASC_TIME)) {
	insert_node = &state->notifications;
}
```

adrubesh:
Thanks. Sorry misunderstood the enum thing at first. Hmm, build says failing but it's compiling fine on my end.

emersion:
Oh, and can you add `sort` to the man page too?

notification.c
```diff
@@ -286,3 +286,56 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+void insert_notification(struct mako_state *state, struct mako_notification *notif) {
+	struct mako_config *config = &state->config;
+	struct wl_list *insert_node;
+
+	if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			!(config->sort_asc & MAKO_SORT_CRITERIA_TIME)) {
+		insert_node = &state->notifications;
+	} else if (config->sort_criteria == MAKO_SORT_CRITERIA_TIME &&
+			(config->sort_asc & MAKO_SORT_CRITERIA_TIME)) {
+		insert_node = &state->notifications;
```
emersion: Hum, this statement has been changed, was there a bug?

adrubesh: Nope. That was just me failing to merge properly. Let me fix.

adrubesh:
Added sort to man page as well. Have not used scdoc before, hope I did it correctly :)

emersion:
Thanks!

PR #41, opened 2018-06-01
Ignore expire_timeout

### LINKED ISSUE #28: Add an option to ignore the client timeout
And use the default one instead

adrubesh:
Fixes #28

### DIFF
--- a/config.c
+++ b/config.c
@@ -185,8 +185,8 @@ static bool apply_config_option(struct mako_config *config, const char *section,
 	} else if (strcmp(name, "default-timeout") == 0) {
 		return parse_int(value, &config->default_timeout);
 	} else if (strcmp(name, "ignore-timeout") == 0) {
-		config->ignore_timeout = true;
-		return true;
+		config->ignore_timeout = strcmp(value, "1") == 0;
+		return config->ignore_timeout || strcmp(value, "0") == 0;
 	} else if (strcmp(name, "output") == 0) {
 		free(config->output);
 		config->output = strdup(value);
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -182,6 +184,9 @@ static bool apply_config_option(struct mako_config *config, const char *section,
 		return parse_int(value, &config->max_visible);
 	} else if (strcmp(name, "default-timeout") == 0) {
 		return parse_int(value, &config->default_timeout);
+	} else if (strcmp(name, "ignore-timeout") == 0) {
+		config->ignore_timeout = true;
```
emersion: Because this can be specified in the config, we should probably parse the value too as a bool (just like `markup`). Also, in the future this will allow things like:

```ini
ignore-timeout=1

[priority=urgent]
ignore-timeout=0
```

mako.1.scd
```diff
@@ -98,6 +98,12 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: 0
 
+*--ignore-timeout*
+	If set, Mako will ignore the expire timeout sent by notifications and use
+	the one provided by _default-timeout_ instead.
+
+	Default: Unset
```
emersion: Should now say: 0

mako.1.scd
```diff
@@ -98,6 +98,12 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: 0
 
+*--ignore-timeout*
+	If set, Mako will ignore the expire timeout sent by notifications and use
```
emersion: Nitpick: I prefer "mako" to "Mako"

PR #42, opened 2018-06-03
Split notification style from config

vilhalmer:
A prerequisite for #3 is to be able to store style information on each notification. This splits the style config bits out into their own struct to make this easier. I also split them up logically in `apply_config_option` to prepare for applying them to different criteria. A lot of this will be rearranged again further down the road, but this way any new features added in the meantime can already be in the right place.

### DIFF
--- a/config.c
+++ b/config.c
@@ -55,23 +55,7 @@ void init_default_style(struct mako_style *style) {
 	style->colors.border = 0x4C7899FF;
 
 	// Everything in the default config is explicitly specified.
-	style->spec = (struct mako_style_spec){
-		.width = true,
-		.height = true,
-		.margin = true,
-		.padding = true,
-		.border_size = true,
-		.font = true,
-		.markup = true,
-		.format = true,
-		.actions = true,
-		.default_timeout = true,
-		.colors = {
-			.background = true,
-			.text = true,
-			.border = true,
-		},
-	};
+	memset(&style->spec, true, sizeof(struct mako_style_spec));
 }
 
 void finish_style(struct mako_style *style) {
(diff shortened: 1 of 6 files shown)


config.c
```diff
@@ -11,42 +11,74 @@
...
+
+	style->actions = true;
+	style->default_timeout = 0;
+
+	style->colors.background = 0x285577FF;
+	style->colors.text = 0xFFFFFFFF;
+	style->colors.border = 0x4C7899FF;
+
+	// Everything in the default config is explicitly specified.
+	style->spec = (struct mako_style_spec){
```
emersion: A shortcut would be:

```c
memset(&style->spec, true, sizeof(struct mako_style_spec));
```

vilhalmer:
Oooh, very nice. Will do.

include/config.h
```diff
@@ -53,6 +74,10 @@ struct mako_config {
 
 void init_config(struct mako_config *config);
```
emersion: Can you rename this to `init_default_config` for consistency?

config.c
```diff
@@ -135,9 +167,32 @@ static bool parse_color(const char *color, uint32_t *out) {
 
 static bool apply_config_option(struct mako_config *config, const char *section,
 		const char *name, const char *value) {
-	if (section != NULL) {
+	// First try to parse this as a global option.
+	if (strcmp(name, "max-visible") == 0) {
```
emersion: I this these should only be parsed if `section == NULL`.

config.c
vilhalmer: Ooh, very nice. Will do.

config.c
vilhalmer: Good catch.

PR #43, opened 2018-06-11
Criteria

### LINKED ISSUE #3: Style differently urgent/low priority notifications

vilhalmer:
At long last, criteria are here! The syntax is as discussed on #3, and is fairly flexible:
```
font=SF Mono 9
background-color=#ffffffff
text-color=#000000ff
margin=40,30

[hidden]
font=monospace 11

[urgency=normal app-name="Test\ App"\\\" actionable=true]
font=monospace 11
text-color=#ff0000ff
background-color=#00000000
margin=10

[!actionable]
margin=0
height=500
background-color=#00ff00ff
markup=0
```

Note that booleans can be specified either as normal, or as bare words (with leading `!` to negate). Quoting, escaping, and escaping within quotes works, the literal name of the app in the above example is `Test App\"`. The backslash within the quotes is unnecessary but doesn't hurt anything. Leaving quotes unmatched or backslashes trailing will raise errors. (I bashed on it a lot, but more testing of the parsing is much appreciated!)

Style options within the global section are stored in an empty criteria (which therefore matches everything). It is always the first in the list, and always initialized with the default style options before parsing the configuration.

Things that can currently be overridden:
- `colors`
- `margin.{top,bottom}` (I have no idea why you would want this.)
- `padding`
- `border_size`
- `font`
- `format`
- `default_timeout`
- `ignore_timeout`
- `markup`

Things that don't work yet:
- `width` (Currently forced to the surface width by `render`.)
- `height` (I think this is just being completely ignored at the moment?)
- `margin.{left,right}` (Need to rework how the surface size is calculated.)
- `actions` (You can set it and it's stored, but turning them off doesn't disable them. See also other caveats below.)

Other caveats:
- I don't know if `markup` and `actions` are going to end up being useful. Their enabled status gets sent as part of the XDG bus setup, which means they can only really be either on or off globally.
    - Disabling `markup` "works" and you see the raw HTML, but since I don't think anyone would ever want this, it might be better to just make it global again. (That way if it's disabled, apps just send plaintext instead.)
    - Disabling `actions` seems slightly more useful. We would want to advertise it as enabled globally if any criteria has it enabled, and then just ignore interactions with notifications where it has been disabled.
- The `[hidden]` section is now fully style-able. However, I'm not happy with the current implementation. I'm going to continue working on a way to make it less of a special case.
- Because styles are generated for each notification at creation, reloading the config will not update the style of existing notifications. It will, however, update the fake notification that shows the number hidden. I have a few ideas about fixing this, but I'm saving it for later.

Closes #3.

### DIFF
--- a/config.c
+++ b/config.c
@@ -16,6 +16,7 @@ void init_default_config(struct mako_config *config) {
 	struct mako_criteria *global_criteria = create_criteria(config);
 	init_default_style(&global_criteria->style);
 
+	init_empty_style(&config->hidden_style);
 	config->hidden_style.format = strdup("(%h more)");
 	config->hidden_style.spec.format = true;
 
@@ -66,6 +67,10 @@ void init_default_style(struct mako_style *style) {
 
 	// Everything in the default config is explicitly specified.
 	memset(&style->spec, true, sizeof(struct mako_style_spec));
+}
+
+void init_empty_style(struct mako_style *style) {
+	memset(style, 0, sizeof(struct mako_style));
 }
 
 void finish_style(struct mako_style *style) {
--- a/config.c
+++ b/config.c
@@ -413,6 +413,11 @@ int load_config_file(struct mako_config *config) {
 		if (line[0] == '[' && line[strlen(line) - 1] == ']') {
 			free(section);
 			section = strndup(line + 1, strlen(line) - 2);
+			if (strcmp(section, "hidden") == 0) {
+				// Skip making a criteria for the hidden section.
+				criteria = NULL;
+				continue;
+			}
 			criteria = create_criteria(config);
 			if (!parse_criteria(section, criteria)) {
 				fprintf(stderr, "[%s:%d] Invalid criteria definition\n", base,
--- a/config.c
+++ b/config.c
@@ -152,89 +152,86 @@ static bool parse_color(const char *color, uint32_t *out) {
 	return true;
 }
 
-static bool apply_config_option(struct mako_config *config, const char *section,
-		const char *name, const char *value) {
-	// First try to parse this as a global option.
-	if (section == NULL) {
-		if (strcmp(name, "max-visible") == 0) {
-			return parse_int(value, &config->max_visible);
-		} else if (strcmp(name, "output") == 0) {
-			free(config->output);
-			config->output = strdup(value);
-			return true;
-		} else if (strcmp(name, "sort") == 0) {
-			if (strcmp(value, "+priority") == 0) {
-				config->sort_criteria |= MAKO_SORT_CRITERIA_URGENCY;
-				config->sort_asc |= MAKO_SORT_CRITERIA_URGENCY;
-			} else if (strcmp(value, "-priority") == 0) {
-				config->sort_criteria |= MAKO_SORT_CRITERIA_URGENCY;
-				config->sort_asc &= ~MAKO_SORT_CRITERIA_URGENCY;
-			} else if (strcmp(value, "+time") == 0) {
-				config->sort_criteria |= MAKO_SORT_CRITERIA_TIME;
-				config->sort_asc |= MAKO_SORT_CRITERIA_TIME;
-			} else if (strcmp(value, "-time") == 0) {
-				config->sort_criteria |= MAKO_SORT_CRITERIA_TIME;
-				config->sort_asc &= ~MAKO_SORT_CRITERIA_TIME;
-			}
-			return true;
-		} else {
-			// We want to try the style options now, so keep going.
+static bool apply_config_option(struct mako_config *config, const char *name,
+		const char *value) {
+	if (strcmp(name, "max-visible") == 0) {
+		return parse_int(value, &config->max_visible);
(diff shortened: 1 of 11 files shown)


config.c
```diff
@@ -59,11 +69,116 @@ void init_default_style(struct mako_style *style) {
...
+}
+
 void finish_style(struct mako_style *style) {
 	free(style->font);
 	free(style->format);
 }
 
+// Update `target` with the values specified in `style`. If a failure occurs,
+// `target` will remain unchanged.
+bool apply_style(struct mako_style *style, struct mako_style *target) {
```
emersion: Generally in C we prefer to have the struct on which the function operates always as the first argument. It seems the comment "`target` will remain unchanged" is outdated.

Also to make it more obvious that an argument isn't mutated, you can make it `const`.

criteria.c
```diff
@@ -0,0 +1,317 @@
...
+
+	if (spec.desktop_entry &&
+			strcmp(criteria->desktop_entry, notif->desktop_entry) != 0) {
+		return false;
+	}
+
+	return true;
+}
+
+bool parse_boolean(const char *string, bool *out) {
```
emersion: This could probably be reused in `config.c` too

criteria.c
```diff
@@ -0,0 +1,317 @@
...
+
+		case MAKO_PARSE_STATE_QUOTE:
+			switch (ch) {
+			case '\\':
+				state = MAKO_PARSE_STATE_QUOTE_ESCAPE;
+				break;
+			case '"':
+				state = MAKO_PARSE_STATE_NORMAL;
+				break;
+			case ' ':
```
emersion: This can be merged with the next `case`

criteria.c
```diff
@@ -0,0 +1,317 @@
...
+	return false;
+}
+
+bool parse_criteria(const char *string, struct mako_criteria *criteria) {
+	// Create space to build up the current token that we're reading. We know
+	// that no single token can ever exceed the length of the entire criteria
+	// string, so that's a safe length to use for the buffer.
+	int token_max_length = strlen(string) + 1;
+	char token[token_max_length];
+	memset(token, 0, token_max_length);
```
emersion: Instead of calling `memset`, you can initialize `token` with `{0}` and all fields will be set to zero.

criteria.c
```diff
@@ -0,0 +1,317 @@
...
+	int token_location = 0;
```
emersion: This can be a `size_t`

criteria.c
```diff
@@ -0,0 +1,317 @@
...
+	struct mako_criteria *criteria =
+		wl_container_of(config->criteria.next, criteria, link);
+	return criteria;
+}
+
+
+// Iterate through `criteria_list`, applying the style from each matching
+// criteria to `notif`. Returns the number of criteria that matched, or -1 if
+// a failure occurs.
+int apply_each_criteria(struct wl_list *criteria_list,
```
emersion: This can return a `ssize_t`

vilhalmer:
Noting this here mostly to hold myself to it: I'm working on updating the manpage to go with this. I can either tack that on to this PR or post another one soon.

criteria.c
vilhalmer: You can't use initialization syntax with a variable-length array. :(

criteria.c
vilhalmer: I was planning on moving this and the existing stuff into a `parse.c` utility-type file in another PR, which would include making `config.c` use it. `config.c` is getting a little bit out of control.

emersion: Yeah, nice idea. So let's keep it that way for now.

criteria.c
emersion: Oh, I didn't know that. RIP

emersion:
>Noting this here mostly to hold myself to it: I'm working on updating the manpage to go with this. I can either tack that on to this PR or post another one soon.

Oh, right, I think this should be done in this PR, so that the manpage is not out-of-date.

emersion:
>Disabling markup "works" and you see the raw HTML, but since I don't think anyone would ever want this, it might be better to just make it global again.

We could strip HTML tags for instance. But yeah, it's not ideal.

>Because styles are generated for each notification at creation, reloading the config will not update the style of existing notifications.

Agreed, this can be filled as a follow-up issue for later.

mako.1.scd
```diff
@@ -124,11 +127,49 @@ Is equivalent to passing *--key=value* to mako from the command line.
...
+any given style option will "win".
+
+The following fields are available in critiera:
+
+- _app-name_ (string)
+- _app-icon_ (string)
+- _urgency_ (one of "low", "normal", "high")
+- _category_ (string)
+- _desktop-entry_ (string)
+- _actionable_ (boolean)
```
emersion: This is missing "hidden"

vilhalmer: Good catch.

emersion:
Thanks, nice work!

PR #49, opened 2018-06-15
Gracefully handle and report failure to reload config

### LINKED ISSUE #44: makoctl reload should fail if parsing config fails

vilhalmer:
- Don't overwrite the config when reloading until it has been fully parsed (woo!)
- Make `reload_config` return a boolean
- Report the success or failure of reload over DBus
- Print a failure message and set the exit code of `makoctl` appropriately

Fixes #44.

### DIFF
--- a/config.c
+++ b/config.c
@@ -531,11 +531,7 @@ bool reload_config(struct mako_config *config) {
 	// currently pointing to local memory instead of the location of the real
 	// criteria struct.
 	wl_list_init(&config->criteria);
-	struct mako_criteria *criteria, *tmp;
-	wl_list_for_each_safe(criteria, tmp, &new_config.criteria, link) {
-		wl_list_remove(&criteria->link);
-		wl_list_insert(config->criteria.prev, &criteria->link);
-	}
+	wl_list_insert_list(&config->criteria, &new_config.criteria);
 
 	return true;
 }
--- a/config.c
+++ b/config.c
@@ -513,9 +513,29 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 	return 0;
 }
 
-void reload_config(struct mako_config *config) {
+bool reload_config(struct mako_config *config) {
+	struct mako_config new_config = {0};
+	init_default_config(&new_config);
+
+	if (load_config_file(&new_config) != 0 ||
+			parse_config_arguments(&new_config, config_argc, config_argv) != 0) {
+		fprintf(stderr, "Failed to reload config\n");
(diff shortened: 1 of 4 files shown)


config.c
```diff
@@ -513,9 +513,29 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
...
+
+	// We have to rebuild the wl_list that contains the criteria, as it is
+	// currently pointing to local memory instead of the location of the real
+	// criteria struct.
+	wl_list_init(&config->criteria);
+	struct mako_criteria *criteria, *tmp;
+	wl_list_for_each_safe(criteria, tmp, &new_config.criteria, link) {
+		wl_list_remove(&criteria->link);
+		wl_list_insert(config->criteria.prev, &criteria->link);
+	}
```
vilhalmer: This is a little gross, but I thought it was worth it to avoid just doing the parsing again. That is an alternative, however.

dbus/mako.c
```diff
@@ -71,18 +71,21 @@ static int handle_reload(sd_bus_message *msg, void *data,
...
+	return sd_bus_reply_method_return(msg, "b", true);
 }
 
 static const sd_bus_vtable service_vtable[] = {
 	SD_BUS_VTABLE_START(0),
 	SD_BUS_METHOD("DismissAllNotifications", "", "", handle_dismiss_all_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("DismissLastNotification", "", "", handle_dismiss_last_notification, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("InvokeAction", "s", "", handle_invoke_action, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("Reload", "", "", handle_reload, SD_BUS_VTABLE_UNPRIVILEGED),
+	SD_BUS_METHOD("Reload", "", "b", handle_reload, SD_BUS_VTABLE_UNPRIVILEGED),
```
emersion: Would it be better to use dbus errors instead of return values?

config.c
emersion: Hmm, I think you can do it with `wl_list_insert_list`

vilhalmer: Good call, I forgot that was there.

dbus/mako.c
vilhalmer: I didn't know those existed, but probably!

emersion:
Thanks!

PR #51, opened 2018-06-15
Clean up notifications before the event loop

### LINKED ISSUE #50: Segfault when terminating mako with visible notifications
To reproduce, `notify-send something` and then immediately `^C` mako.
```
Jun 14 21:50:26 sayaka kernel: mako[19007]: segfault at 8 ip 00007f92b4543947 sp 00007ffe0c8c1b28 error 6 in libwayland-client.so.0.3.0[7f92b453a000+d000]
Jun 14 21:50:26 sayaka systemd[1]: Started Process Core Dump (PID 19035/UID 0).
Jun 14 21:50:26 sayaka systemd-coredump[19038]: Process 19007 (mako) of user 1000 dumped core.
                                                
    Stack trace of thread 19007:
    #0  0x00007f92b4543947 wl_list_remove (libwayland-client.so.0)
    #1  0x000055a891f5cde1 n/a (mako)
    #2  0x000055a891f5dec5 n/a (mako)
    #3  0x000055a891f5b816 n/a (mako)
    #4  0x00007f92b41a106b __libc_start_main (libc.so.6)
    #5  0x000055a891f5b8aa n/a (mako)
```

vilhalmer:
This allows the notifications to destroy their timers. Previously, all of the timers would be destroyed by `finish_event_loop`, then `finish_notification` would attempt to re-destroy them and become rather unhappy.

Fixes #50.

### DIFF
--- a/main.c
+++ b/main.c
@@ -49,13 +49,13 @@ static bool init(struct mako_state *state) {
 }
 
 static void finish(struct mako_state *state) {
-	finish_event_loop(&state->event_loop);
-	finish_wayland(state);
-	finish_dbus(state);
 	struct mako_notification *notif, *tmp;
 	wl_list_for_each_safe(notif, tmp, &state->notifications, link) {
 		destroy_notification(notif);
 	}
+	finish_event_loop(&state->event_loop);
+	finish_wayland(state);
+	finish_dbus(state);
 }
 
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #53, opened 2018-06-18
Rebuild notification styles on reload

### LINKED ISSUE #45: Reloading the config doesn't update current notifications style
See #43

vilhalmer:
Fixes #45. I actually started working on this before deciding that it depended on #44, and that one turned out to be the harder one.


emersion:
Thanks!

PR #56, opened 2018-07-03
Add 'expiring' criteria

vilhalmer:
A criteria to match whether a notification will ever disappear on its own.

This allows you to set `ignore-timeout` + `default-timeout` on non-expiring notifications in general, rather than having to match on applications that tend to send them. This effectively prevents non-expiring notifications entirely:
```
[!expiring]
ignore-timeout=true
default-timeout=5000
```
Obviously you can do other things with it as well, but this is the most immediately useful application.

This also adds a field to `mako_notification` to track the requested timeout, as opposed to what we decided to give it. This was necessary for the value to be present at the time criteria are applied. This means that you can only match based on the timeout that the notifier requested, not what you have overridden it to be in other criteria sections using `ignore-timeout`. There's no way to "fix" this without introducing multiple stages of criteria matching, which is a road I really don't want to go down. This case is a bit unique, as `ignore-timeout` is the only option which has the effect of changing something provided by the notifier.

I don't see this as a big issue, as if you are overriding the timeout for something specific, you can also just set the other style options in that criteria section.

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -207,11 +207,12 @@ static int handle_notify(sd_bus_message *msg, void *data,
 		return ret;
 	}
 
-	int32_t expire_timeout;
-	ret = sd_bus_message_read(msg, "i", &expire_timeout);
+	int32_t requested_timeout;
+	ret = sd_bus_message_read(msg, "i", &requested_timeout);
 	if (ret < 0) {
 		return ret;
 	}
+	notif->requested_timeout = requested_timeout;
 
 	int match_count = apply_each_criteria(&state->config.criteria, notif);
 	if (match_count == -1) {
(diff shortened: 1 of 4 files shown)


dbus/xdg.c
```diff
@@ -227,7 +228,8 @@ static int handle_notify(sd_bus_message *msg, void *data,
 		return -1;
 	}
 
-	if (expire_timeout < 0 || notif->style.ignore_timeout) {
+	int32_t expire_timeout;
+	if (notif->requested_timeout < 0 || notif->style.ignore_timeout) {
```
emersion: If this condition isn't satisfied, `expire_timeout` will be uninitialized. We should probably set it to `requested_timeout`.

vilhalmer: Fixed, good catch. My test notification wasn't specifying one.

PR #57, opened 2018-07-16
Introduce anchor option

RyanDwyer:
This adds an `anchor` option, to allow the user to choose which corner of the output the notifications should appear in.

Usage:

`--anchor topright|bottomright|bottomleft|topleft`

### DIFF
--- a/config.c
+++ b/config.c
@@ -208,16 +208,16 @@ static bool apply_config_option(struct mako_config *config, const char *name,
 		}
 		return true;
 	} else if (strcmp(name, "anchor") == 0) {
-		if (strcmp(value, "topright") == 0) {
+		if (strcmp(value, "top-right") == 0) {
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -414,7 +436,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 
 		const char *name = long_options[option_index].name;
 		if (!apply_style_option(&global_criteria(config)->style, name, optarg)
-				|| apply_config_option(config, name, optarg)) {
+				&& !apply_config_option(config, name, optarg)) {
```
RyanDwyer: This looked wrong to me, and I was  getting a parse failure for `--output` until I changed it to this.

emersion: Oh right, good catch

emersion:
Sweet, thanks!

PR #59, opened 2018-08-16
Fix memory leaks

ianyfan:
Except I'm not sure if the loop in the second commit is necessarily in the right place (or if the comment is correct, i.e. that I'm understanding the problem properly)

### DIFF
--- a/dbus/dbus.c
+++ b/dbus/dbus.c
@@ -45,5 +45,5 @@ error:
 void finish_dbus(struct mako_state *state) {
 	sd_bus_slot_unref(state->xdg_slot);
 	sd_bus_slot_unref(state->mako_slot);
-	sd_bus_unref(state->bus);
+	sd_bus_flush_close_unref(state->bus);
 }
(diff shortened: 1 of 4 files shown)


dbus/dbus.c
```diff
@@ -43,6 +43,9 @@ bool init_dbus(struct mako_state *state) {
 }
 
 void finish_dbus(struct mako_state *state) {
+	while ((sd_bus_process(state->bus, NULL)) > 0) {
```
emersion: According to the [docs](https://www.freedesktop.org/software/systemd/man/sd_bus_default.html), it's a good idea to call `sd_bus_flush` followed by `sd_bus_close` before dropping the ref (the former will flush outgoing messages, the latter drop incoming messages).

ianyfan: Found a better way, I think: https://github.com/systemd/systemd/pull/485

Annoyingly undocumented though.

emersion:
Thanks!

PR #62, opened 2018-09-06
add support for [\n] character format parameter inside config file

### LINKED ISSUE #61: Changing fromat from configuration file does not recognize '\n' character
As an example: if you put`format=%s\n%b` in the config file, the notification result wold be   `Summary\nBody` (those are two value that i'm using for testing).
Debugging the line that mako gets from the config file, you get that the `\` char is escaped.
gdb output (breakpoint at config.c:378):
```
(gdb) p eq + 1
$1 = 0x555555572387 "%s\\n%b"
```

masozzi:
This fixes #61 
As a consequence also the char `\` needs to be escaped.

i'll update the `mako.1.sdc` file once the modification are approved

### DIFF
--- a/types.c
+++ b/types.c
@@ -126,16 +126,15 @@ bool parse_directional(const char *string, struct mako_directional *out) {
 }
 
 bool parse_format(const char *string, char **out) {
-	int token_max_length = strlen(string) + 1;
+	size_t token_max_length = strlen(string) + 1;
 	char token[token_max_length];
 	memset(token, 0, token_max_length);
 	size_t token_location = 0;
 
 	enum mako_parse_state state = MAKO_PARSE_STATE_NORMAL;
-	const char *location = string;
+	for (size_t i = 0; i < token_max_length; ++i) {
+		char ch = string[i];
 
-	char ch;
-	while ((ch = *location++) != '\0') {
 		switch (state) {
 		case MAKO_PARSE_STATE_ESCAPE:
 			switch (ch) {
@@ -145,7 +144,12 @@ bool parse_format(const char *string, char **out) {
 				break;
 
 			case '\\':
+				token[token_location] = '\\';
+				++token_location;
+				break;
+
 			default:
+				++token_location;
 				token[token_location] = ch;
 				++token_location;
 				break;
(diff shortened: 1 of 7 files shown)


types.c
```diff
@@ -123,3 +124,55 @@ bool parse_directional(const char *string, struct mako_directional *out) {
...
+	memset(token, 0, token_max_length);
+	size_t token_location = 0;
+
+	enum mako_parse_state state = MAKO_PARSE_STATE_NORMAL;
+	const char *location = string;
+
+	char ch;
+	while ((ch = *location++) != '\0') {
+		switch (state) {
+			case MAKO_PARSE_STATE_ESCAPE:
```
ascent12: I think the style is to not indent case labels.

types.c
```diff
@@ -123,3 +124,55 @@ bool parse_directional(const char *string, struct mako_directional *out) {
 	free(components);
 	return true;
 }
+
+bool parse_format(const char *string, char **out) {
+	int token_max_length = strlen(string) + 1;
```
ascent12: size_t is more correct here.

types.c
```diff
@@ -123,3 +124,55 @@ bool parse_directional(const char *string, struct mako_directional *out) {
...
+	char token[token_max_length];
+	memset(token, 0, token_max_length);
+	size_t token_location = 0;
+
+	enum mako_parse_state state = MAKO_PARSE_STATE_NORMAL;
+	const char *location = string;
+
+	char ch;
+	while ((ch = *location++) != '\0') {
```
ascent12: I think a normal for loop would be better.

types.c
```diff
@@ -123,3 +124,55 @@ bool parse_directional(const char *string, struct mako_directional *out) {
...
+		switch (state) {
+			case MAKO_PARSE_STATE_ESCAPE:
+				switch (ch) {
+					case 'n':
+						token[token_location] = '\n';
+						++token_location;
+						break;
+
+					case '\\':
+					default:
```
ascent12: I'm not sure if you want to ignore unknown escape sequences or not. Sometimes that can mess with the config file's forwards compatibility.
That's up to @emersion though.

emersion:
Thanks for your PR!

I agree with @ascent12's comments (thanks!).

>I'm not sure if you want to ignore unknown escape sequences or not. Sometimes that can mess with the config file's forwards compatibility.

I'm not super-concerned about this. Maybe keeping unknown escaped sequences as-is would be a little bit better.

emersion:
Thanks!

masozzi:
sorry, i should have pinged you.
by the way, thanks for the help!

PR #65, opened 2018-09-24
fix: add default in switch

meakio:
Thanks for the solution. I don't know how to don't pollute the git so I create a new pull request.


render.c
```diff
@@ -51,6 +52,8 @@ static cairo_subpixel_order_t get_cairo_subpixel_order(
 		return CAIRO_SUBPIXEL_ORDER_VRGB;
 	case WL_OUTPUT_SUBPIXEL_VERTICAL_BGR:
 		return CAIRO_SUBPIXEL_ORDER_VBGR;
+	default:
+		assert(0);
```
emersion: Does it fix your problem if you put the `assert(0)` outside of the `switch`?

```c
switch (…) {
case …:
    return …
}
assert(0);
```

You can edit your latest commit by doing:

```
git commit -a --amend
git push --force
```

meakio: It does solve my compilation error.
Thanks for the git tip.

Issue #66, opened 2018-09-25
Add option to group notifications with same title


cauebs:
I'm using Telegram Desktop and, when I receive several messages, the notifications pile up even if they are from the same person. There could be a way to group them and show the title (is it called summary?) only once, at the top.

emersion:
Yeah. Dunst has a [setting to stack notifications which are exactly the same](https://github.com/dunst-project/dunst/commit/bf6c1a6406ad28899dec11504e3de87125e526d6), this feature would be a superset of that.

vilhalmer:
I'm imagining configuring this using a list of criteria that must match, something like:
```
stack=app-name,summary,body  # dunst-like for most

[app-name=Telegram]
stack=summary  # group by conversation

[urgency=high]
stack=none
```
Don't know if we want to try to handle the OR case. Probably not. Maybe it should be something other than a comma to make it more obvious that it's AND.

Do we want to be fancy with this and allow them to be expanded somehow? Or are we ok with just erasing all but the top one and keeping track of a count, like dunst does?

vilhalmer:
I'm starting to work on the design for this feature and I have a couple more questions:
- How should this interact with `sort-criteria`? Do we want stacking to always "win" and position the entire stack where the first notification in the stack would end up? Or do we sort first and then only stack ones that are adjacent to each other? (I am leaning toward the former, to increase the odds that stacking will happen.)
- Do we want to call this stacking, grouping, or some other third thing?

I think I've managed to fold both of the options in my previous question (keep all vs. only the top one) into the design using some new criteria, and it should also allow us to replace the current special-case implementation of `hidden` as well. I'm not going to share yet, until I know it works. :)

cauebs:
Since this is going to be used almost exclusively by "power users", I think the best approach is to allow these things to be configurable.
About the naming, I think it's mostly bikeshedding, but I'd prefer "grouping".

emersion:
>How should this interact with sort-criteria? Do we want stacking to always "win" and position the entire stack where the first notification in the stack would end up? Or do we sort first and then only stack ones that are adjacent to each other? (I am leaning toward the former, to increase the odds that stacking will happen.)

I agree with you: first grouping should happen, and then sorting.

>About the naming, I think it's mostly bikeshedding, but I'd prefer "grouping".

Same.

>and it should also allow us to replace the current special-case implementation of hidden as well.

Yay!

0xdec:
[...]

Issue #55, opened 2018-07-03
Receiving notification content with invalid pango markup breaks format markup


vilhalmer:
If the body of a notification contains something that pango considers to be invalid markup, the markup within the user-specified format will be ignored as well (see screenshot for an example). This is because we're parsing the entire formatted notification at once. This is required to have relative markup work correctly, but it could be made more robust. Here's what I propose:

- Attempt to parse the body alone using `pango_parse_markup`, without saving the result.
- If this reported success, continue as usual.
- Otherwise, escape the body using the existing `escape_markup` function before formatting.
- Parse the entire thing as we do now.

This will allow the user's format to continue working (e.g. to display the summary in bold) even when the body is unparseable. It's arguable that this isn't a mako bug, but the result of clients sending notifications that don't follow the spec. However, I've now seen this from two separate programs (including Slack), so it's probably worth putting a bandaid on.

---

![screenshot_2018-07-03-131047](https://user-images.githubusercontent.com/201439/42234477-adb621ae-7ec2-11e8-9902-926d21626643.png)
An example of the `<` character causing parsing to fail.
```
cannot parse pango markup: Error on line 2 char 7: '<' is not a valid name
```

emersion:
Hmm, yes, I've experienced this too. The spec allows us to advertise that we support markup, but doesn't allow clients to opt-out. :(

I think the fix you suggest is reasonable.

progandy:
Mako should probably filter unsupported html tags as well and only keep those that are defined in the freedesktop specification.

vilhalmer:
I don't know if there's a good way to do this without accidentally ruining some notifications' content. For example, if someone sends me a message on slack that contains a tag, I don't want that tag to be removed, since it was intended as the literal text of the tag.

progandy:
I guess in that case keep the tags for now and when a bug report is filed, add a criteria option to strip tags when receiving notifications from known bad clients.

Issue #48, opened 2018-06-14
Horizontal margins cannot be overridden in criteria


vilhalmer:
#43.

Issue #47, opened 2018-06-14
Disabling actions with criteria has no effect


vilhalmer:
#43.

PR #68, opened 2018-09-30
Add apply_superset_style to calculate the maximum values for all criteria styles

vilhalmer:
This replaces `global_criteria` as the way to find out about notification styles in the general case, rather than for a specific notification. This is useful for two things that aren't currently possible to do accurately:
- Sizing the layer surface (what's the widest notification we can display?)
- Figuring out which capabilities to advertise (are they set for at least one criteria?)

This doesn't directly fix any issues, but was necessary for #46, #47, and #48, and will improve the fix for #55.

The main behavior change here is that we're now advertising DBus capabilities based on whether any of the criteria are using them, rather than just the default section. This means that you can now, for example, turn off markup globally but reenable it for a specific type of notification. Before, you could only do the opposite.

I've eliminated all but one use of `global_criteria`. It will be able to go away entirely after #66, because the stacking mechanism can be used to make the `hidden` criteria no longer a special case (yay!).

The one tricky part of this was the "maximum" value for `format`: a collection of all of the specifiers that appear in at least one style. I added format validation to make it a little easier (by knowing what the maximum length of that collection can be).

Finally, I took the opportunity of needing to generate the "superstyle" on each reconfig to unify config loading during startup with that of reload_config. The `argc` and `argv` are now stored on the `mako_state` to make that easier.


config.c
```diff
@@ -184,6 +193,85 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
...
+	target->spec.margin = true;
+	target->spec.padding = true;
+	target->spec.border_size = true;
+	target->spec.default_timeout = true;
+	target->spec.markup = true;
+	target->spec.actions = true;
+	target->spec.format = true;
+
+	if (target->format) {
+		free(target->format);
```
emersion: No need for these `if`s. `free(NULL)` is valid and does nothing.

criteria.c
```diff
@@ -134,6 +135,9 @@ bool parse_criteria(const char *string, struct mako_criteria *criteria) {
 				++token_location;
 			}
 			break;
+		default:
```
emersion: Hmm. I'm not a fan of adding `default` cases since most compilers will emit a warning when an enum case isn't handled.

include/types.h
```diff
@@ -25,6 +25,10 @@ struct mako_directional {
 };
 
 bool parse_directional(const char *string, struct mako_directional *out);
+
+// List of specifier characters that can appear in a format string.
+static const char VALID_FORMAT_SPECIFIERS[] = "%asbht";
```
emersion: I'd like not to introduce `static` definitions in headers. Instead we can declare this constant as `extern` here and define it in a `.c` file.

emersion:
Thanks!

PR #74, opened 2018-10-13
Allow overriding horizontal margins in criteria

### LINKED ISSUE #48: Horizontal margins cannot be overridden in criteria

vilhalmer:
Fixes #48. As we had discussed at some point in the distant past, I had to remove the margin from around the layer surface for this.

I also unified the calls to `zwlr_layer_surface_v1_set_size` during `send_frame`. I couldn't find any situations where the early return when unconfigured actually happened, but if you had one in mind when you added it originally I can put it back.

### DIFF
--- a/render.c
+++ b/render.c
@@ -79,10 +79,13 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	int notif_width =
 		(style->width <= state->width) ? style->width : state->width;
 
-	// Calculate the appropriate offset if we're right-aligned.
-	bool right_align =
-		(state->config.anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT);
-	int offset_x = right_align ? (state->width - notif_width) : 0;
+	int offset_x;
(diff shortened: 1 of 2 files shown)


emersion:
>I couldn't find any situations where the early return when unconfigured actually happened

This happens each time the height changes, ie. each time a notification is added or removed (except the first one).

render.c
```diff
@@ -79,10 +79,13 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
...
 
-	// Calculate the appropriate offset if we're right-aligned.
-	bool right_align =
-		(state->config.anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT);
-	int offset_x = right_align ? (state->width - notif_width) : 0;
+	int offset_x;
+	if (state->config.anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT) {
+		offset_x = state->width - notif_width - style->margin.right;
+	}
+	else {
```
emersion: Style: `}` and `else {` on the same line

vilhalmer: The one style thing I can never adjust to :)

emersion:
Thanks!

PR #76, opened 2018-10-14
Improve body parsing

### LINKED ISSUE #55: Receiving notification content with invalid pango markup breaks format markup

vilhalmer:
When receiving invalid markup in a notification, escape it even if markup is enabled. This prevents the bad markup from breaking the user's defined `format`. Additionally, the `markup` flag now only applies to application-provided content (i.e., the body). If a user doesn't want markup in their style, they should change `format` anyway since the default is quite ugly without markup enabled.

Fixes #55.

---
screenshot_2018-10-14-125128
A notification containing invalid markup which still formats correctly.

screenshot_2018-10-14-125308
A notification containing valid markup, but with `markup=false`.

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -187,7 +187,7 @@ char *format_notif_text(char variable, bool *markup, void *data) {
 	case 's':
 		return strdup(notif->summary);
 	case 'b':
-		*markup = true;
+		*markup = notif->style.markup;
 		return strdup(notif->body);
 	}
 	return NULL;
--- a/notification.c
+++ b/notification.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <pango/pangocairo.h>
 #include <wayland-client.h>
 #ifdef __linux__
 #include <linux/input-event-codes.h>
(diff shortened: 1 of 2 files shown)


emersion:
Thanks!

PR #77, opened 2018-10-20
[WIP] Add support for elogind

DanySpin97:
`busctl` is not shipped with elogind so makoctl does not work.


emersion:
Thanks!

emersion:
Note that I'd like to have a library that just ships `sd-bus`.

PR #79, opened 2018-10-21
 event-loop: poll a signalfd to know when to quit

CameronNemo:
Signed-off-by: Cameron Nemo <camerontnorman@gmail.com>

### DIFF
--- a/main.c
+++ b/main.c
@@ -1,6 +1,5 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
-#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


emersion:
On one hand signalfd is Linux-specific, on the other hand we already use timerfd and epoll-shim already has support for both timerfd and signalfd.

main.c
```diff
@@ -37,6 +38,23 @@ static const char usage[] =
 	"\n"
 	"Colors can be specified with the format #RRGGBB or #RRGGBBAA.\n";
 
+static bool init_signalfd(struct mako_state *state) {
```
emersion: I think it would make more sense to move this into `event-loop.c`

emersion:
Thanks!

PR #83, opened 2018-10-21
Apply bottom margin to last notification

### LINKED ISSUE #82: Bottom margin not added to bottom item when bottom-anchored
In other words, `bottom-left` and `bottom-right` results in the notifications being glued to the top of the swaybar, regardless of the bottom margin. This isn't the case for the top. Happens in the current version, but not in 1.1. Probably #74?

vilhalmer:
Fixes #82.


emersion:
Thanks!

jeremejevs:
Damn, that's fast :slightly_smiling_face: Thank you, works!

Issue #72, opened 2018-10-13
reserved identifier violation


elfring:
I would like to point out that an identifier like “[`_MAKO_CRITERIA_H`](https://github.com/emersion/mako/blob/2690fd68cb2adb6471ebcd68fc131ad128154176/include/criteria.h#L1 "Update candidate")” [does eventually not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

emersion:
Sure

PR #97, opened 2018-11-24
Don't use all-caps identifiers with a leading underscore

emersion:
Fixes https://github.com/emersion/mako/issues/72


elfring:
Thanks for your adjustment of affected include guards.

PR #98, opened 2018-11-30
Grouping

vilhalmer:
Alright, part one of #66! This is not a complete solution, but I didn't want to stay silent any longer and this seemed like a good first chunk. This PR adds the following:

- Matching on the exact contents of summary and body
- Specify which criteria fields should be used for grouping notifications
- Group them when adding or removing notifications
- Re-apply criteria to all notifications that are affected by the grouping

You can use it something like this:
```
[app-name=Firefox]
group=summary
```

I was originally going to go down the path of having a first-class group type which would contain one or more notifications, and replace the list of notifications with groups on `mako_state`. This made interacting with individual notifications such a pain that I decided it would be cleaner to just re-order the list based on the grouping rules whenever it was modified.

The grouping specification reuses `mako_criteria_spec`, I've added a function to parse them from a  comma separated string. There's also now a function to create a criteria that matches a given notification, which is then used to compare them when attempting grouping. This includes matching on summary and body, but I didn't expose them as criteria because they still don't seem super useful to the user without regex support.

The current default setting is to group notifications that are identical in every way. There is no way to specify "don't group at all", but I'm planning to add that in the next chunk. I'm trying to figure out if there's a way to do it without adding a special field to `mako_criteria_spec` (since an empty spec would match everything).

As it is, when a group is created the notifications will all be pulled together within the list based on whatever sorting mode is set, but will still appear as normal notifications. The next step will be to actually allow visually combining grouped notifications. This will be accomplished by adding some new criteria that select based on a notification's index within a group (if there is one). The default will probably be to simply make the height and margins of all but the first in the group zero, and to add the number of notifications in the group to the format. Adding these criteria is why it's necessary to re-apply criteria to any newly-grouped notifications. I was hoping to avoid doing an extra criteria match, but there was no clean way to do so. Only doing it to notifications in the group seems like a reasonable trade.

### DIFF
--- a/config.c
+++ b/config.c
@@ -84,6 +84,9 @@ void init_default_style(struct mako_style *style) {
 	style->colors.text = 0xFFFFFFFF;
 	style->colors.border = 0x4C7899FF;
 
+	// Only completely identical notifications should group by default.
+	memset(&style->group_criteria_spec, true, sizeof(struct mako_criteria_spec));
+
 	// Everything in the default config is explicitly specified.
 	memset(&style->spec, true, sizeof(struct mako_style_spec));
 }
@@ -194,6 +197,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	if (style->spec.colors.border) {
 		target->colors.border = style->colors.border;
 		target->spec.colors.border = true;
+	}
+
+	if (style->spec.group_criteria_spec) {
+		target->group_criteria_spec = style->group_criteria_spec;
+		target->spec.group_criteria_spec = true;
 	}
 
 	return true;
@@ -381,6 +389,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	} else if (strcmp(name, "ignore-timeout") == 0) {
 		return spec->ignore_timeout =
 			parse_boolean(value, &style->ignore_timeout);
+	} else if (strcmp(name, "group") == 0) {
+		return spec->group_criteria_spec =
+			parse_criteria_spec(value, &style->group_criteria_spec);
 	}
 
 	return false;
--- a/criteria.c
+++ b/criteria.c
@@ -177,46 +177,6 @@ bool parse_criteria(const char *string, struct mako_criteria *criteria) {
 	return true;
 }
 
-bool parse_criteria_spec(const char *string, struct mako_criteria_spec *out) {
-	// Clear any existing specified fields in the output spec.
-	memset(out, 0, sizeof(struct mako_criteria_spec));
-
-	char *components = strdup(string);
-	char *saveptr = NULL;
-	char *token = strtok_r(components, ",", &saveptr);
-
-	while (token) {
-		// Can't just use &= because then we nave no way to report invalid
-		// values. :(
-		if (strcmp(token, "app-name") == 0) {
-			out->app_name = true;
-		} else if (strcmp(token, "app-icon") == 0) {
-			out->app_icon = true;
-		} else if (strcmp(token, "actionable") == 0) {
-			out->actionable = true;
-		} else if (strcmp(token, "expiring") == 0) {
-			out->expiring = true;
-		} else if (strcmp(token, "urgency") == 0) {
-			out->urgency = true;
-		} else if (strcmp(token, "category") == 0) {
-			out->category = true;
-		} else if (strcmp(token, "desktop-entry") == 0) {
-			out->desktop_entry = true;
-		} else if (strcmp(token, "summary") == 0) {
-			out->summary = true;
-		} else if (strcmp(token, "body") == 0) {
-			out->body = true;
-		} else {
-			fprintf(stderr, "Unknown criteria field '%s'\n", token);
-			return false;
-		}
-
-		token = strtok_r(NULL, ",", &saveptr);
-	}
-
-	return true;
-}
-
 // Takes a token from the criteria string that looks like "key=value", figures
 // out which field of the criteria "key" refers to, and sets it to "value".
 // Any further equal signs are assumed to be part of the value. If there is no .
--- a/criteria.c
+++ b/criteria.c
@@ -31,6 +31,8 @@ void destroy_criteria(struct mako_criteria *criteria) {
 	free(criteria->app_icon);
 	free(criteria->category);
 	free(criteria->desktop_entry);
+	free(criteria->summary);
+	free(criteria->body);
 	free(criteria);
 }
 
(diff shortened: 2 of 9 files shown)


emersion:
Eh, I somehow completely forgot about this. Sorry, I'll review this in the next few days. Next time feel free to nag me. :)

vilhalmer:
Haha, I was in no hurry because I've been too busy to do the next part anyway.

emersion:
This PR looks really great! Though, I'm kind of worried about the code complexity. I really don't want the codebase to grow into something too complex.

>I was originally going to go down the path of having a first-class group type which would contain one or more notifications, and replace the list of notifications with groups on mako_state. This made interacting with individual notifications such a pain that I decided it would be cleaner to just re-order the list based on the grouping rules whenever it was modified.

Hmm. To be honest I'm not sure this will work well when we'll want to render. Any plans?

>There is no way to specify "don't group at all", but I'm planning to add that in the next chunk.

No grouping should probably be the default.

>I'm trying to figure out if there's a way to do it without adding a special field to mako_criteria_spec (since an empty spec would match everything).

Maybe a NULL spec? Maybe this blows everything up. :D

It would be easy to add a new `parse_group_spec` function that would call `parse_criteria_spec` except when the input is exactly something like `none`.

>This will be accomplished by adding some new criteria that select based on a notification's index within a group (if there is one).

Hmm. How does dunst handle this? Is something like "first"/"not first" good enough?

>The default will probably be to simply make the height and margins of all but the first in the group zero

Eheh, maybe we could introduce a "hidden" style option, not sure.

> and to add the number of notifications in the group to the format.

People will probably want to only show the number if there's more than one notification in the group. Not sure how this should look if the count is in e.g. parentheses.

vilhalmer:
> This PR looks really great! Though, I'm kind of worried about the code complexity. I really don't want the codebase to grow into something too complex.

I was really trying to avoid needing to do the re-matching of the criteria, but I couldn't come up with a way to do it in a single pass because we don't know what to group on until we've already completed the first match. I definitely want to avoid making it unbounded though, two passes should always be sufficient. I'll keep thinking about this too, maybe there is a way to simplify it.

> Hmm. To be honest I'm not sure this will work well when we'll want to render. Any plans?

I don't think that render needs to take into account grouping at all, but instead we can just change the style of the grouped notifications to make them appear visually as a group when rendered. More about this toward the bottom.

> No grouping should probably be the default.

I agree.

> Maybe a NULL spec? Maybe this blows everything up. :D
> 
> It would be easy to add a new `parse_group_spec` function that would call `parse_criteria_spec` except when the input is exactly something like `none`.

It's probably best to just add a special criteria like `none` that never matches any notifications. Then the comparison logic is "special", but the way it's stored on the criteria spec isn't.

While I was thinking about this reply I had an alternate idea: add a criteria to match on the notification's index. Then it's not a special case, but is still guaranteed to never match. It would be otherwise kind of a useless criteria, but we don't even really need to advertise the fact that it's there. If we ever expanded the number of operators for comparing criteria you could use it to do something like make your notifications alternate between a few colors, but that seems like a stretch.

> > This will be accomplished by adding some new criteria that select based on a notification's index within a group (if there is one).
> 
> Hmm. How does dunst handle this? Is something like "first"/"not first" good enough?

> People will probably want to only show the number if there's more than one notification in the group. Not sure how this should look if the count is in e.g. parentheses.

I am picturing this all being handled by inserting criteria into the chain right after the global one that do something like:
```
[group-index=1]
format=(%g) %s\n%b

[group-index>1]
height=0
margin=0
```

Which effectively makes only the top notification in a group visible, and adds the count to the format only when grouped (there's no such thing as a single-notification group, in that case the group index is 0 to make them distinguishable). These criteria are always present and invisible to the user but, because the last criteria to match always wins, configuring grouping is just a matter of adding the same criteria to your config and changing the values.

If we don't want to add the new greater-than operator for comparison, we could instead just add one like `grouped` that means the same thing (and perhaps change `group-index=1` to `top` or something). I don't feel too strongly one way or the other about that.

I originally wanted to do it like this:
```
[group-index]
height=0

[group-index=1]
height=<whatever>
```
to collapse them all and just reset the height of the top one, but there's no way to figure out what the height should be without making the user always configure it in two spots, which would be terrible.

> maybe we could introduce a "hidden" style option

I think this is a good idea, and it would also allow the "reset" method above to work, since the height wouldn't be lost.

Now for why I chose this method: I had the idea that it would be neat to be able to play with how groups looked without needing to add any additional new features. Setting up the criteria this way allows you to do interesting stuff like this, for example:
```
width=200
background-color=#eeeeee

[group-index=2]
format=\n
width=180
height=10
background-color=#ededed

[group-index=3]
format=\n
width=160
height=10
background-color=#ececec
```
which gives the visual effect of an actual stack of notifications similar to the style of iOS 12 (these exact values probably wouldn't actually look good but you get the idea). It even reacts to there being only two in the stack.

In less crazy land, you could do something as simple as just deciding to display the top two notifications in each group, or the top three (assuming we add the greater-than operator, otherwise it's a bit less nice).

Anyway, I digress. I thought it would be neat to see what people could come up with using the relatively small tool set we've already given them, and (hopefully) without increasing the complexity too much in the process. If you want to veto this though, I totally understand because it is very much an extra on top of the feature request. That's why I put up this section separately. :)

Regardless of whether you want to go with this method or not, I've been thinking about whether there's a way to have groups be a first-class construct without also increasing the difficulty of working with the notifications. The best idea I've come up with so far is keeping the global list of notifications, but allowing you to walk the groups by having a separate linked list with the first of each group in it. Does that seem reasonable? I think it would at least reduce the amount of iterating we have to do. It's been a bit hard to decide what to optimize for, but I think overall you're more likely to have a lot of ungrouped notifications than a group with a ton in it.

...sorry about the brain dump, heh.

emersion:
>I don't think that render needs to take into account grouping at all, but instead we can just change the style of the grouped notifications to make them appear visually as a group when rendered.

Makes sense.

>While I was thinking about this reply I had an alternate idea: add a criteria to match on the notification's index.

Hmm, I think I still prefer the other approach, this one seems a little bit convoluted.

>I am picturing this all being handled by inserting criteria into the chain right after the global one that do something like

Okay, looks good.

>group-index

I'd either make these start at zero, or rename to something else. Indexes starting at 1 are confusing.

>which gives the visual effect of an actual stack of notifications similar to the style of iOS 12 (these exact values probably wouldn't actually look good but you get the idea).

This doesn't play well with borders, but is a, let's say, interesting hack.

>In less crazy land, you could do something as simple as just deciding to display the top two notifications in each group, or the top three (assuming we add the greater-than operator, otherwise it's a bit less nice).

Yeah, this seems useful.

> The best idea I've come up with so far is keeping the global list of notifications, but allowing you to walk the groups by having a separate linked list with the first of each group in it. Does that seem reasonable?

As discussed on IRC, we can probably have a "group index" field in `mako_notification`. We can re-compute it each time the notification list changes. I don't know if having an explicit list would allow us to simplify anything.

emersion:
Thanks!

PR #106, opened 2019-01-06
Use wl_surface.damage_buffer

emersion:
Per [1], it's better to use damage_buffer instead of damage.

[1]: https://gitlab.freedesktop.org/wayland/wayland/commit/921d0548035673a1bf6aeb9396b9bc728133411e


vilhalmer:
LGTM, mind rebasing?

vilhalmer:
Thanks!

PR #109, opened 2019-02-07
Add progress indicator

moverest:
I found it useful to show a progress indicator in a visual way in a notification so I've implemented it.

notif

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -198,6 +198,13 @@ static int handle_notify(sd_bus_message *msg, void *data,
 			}
 			free(notif->desktop_entry);
 			notif->desktop_entry = strdup(desktop_entry);
+		} else if (strcmp(hint, "value") == 0) {
+			int32_t progress = 0;
+			ret = sd_bus_message_read(msg, "v", "i", &progress);
+			if (ret < 0) {
+				return ret;
+			}
+			notif->progress = progress;
 		} else {
 			ret = sd_bus_message_skip(msg, "v");
 			if (ret < 0) {
--- a/main.c
+++ b/main.c
@@ -28,6 +28,7 @@ static const char usage[] =
 	"                                  meaning as in CSS.\n"
 	"      --border-size <px>          Border size.\n"
 	"      --border-color <color>      Border color.\n"
+	"      --progress-color <color>    Progress indicator color.\n"
 	"      --markup <0|1>              Enable/disable markup.\n"
 	"      --format <format>           Format string.\n"
 	"      --hidden-format <format>    Format string.\n"
--- a/mako.1.scd
+++ b/mako.1.scd
@@ -107,6 +107,12 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: #4C7899FF
 
+*--progress-color* _color_
+	Set popup progress indicator color to _color_. See *COLOR* for more
+	information.
+
+	Default: #5588AAFF
+
 *--markup* 0|1
 	If 1, enable Pango markup. If 0, disable Pango markup. If enabled, Pango
 	markup will be interpreted in your format specifier and in the body of
--- a/mako.1.scd
+++ b/mako.1.scd
@@ -107,11 +107,14 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: #4C7899FF
 
-*--progress-color* _color_
+*--progress-color* [over|source] _color_
 	Set popup progress indicator color to _color_. See *COLOR* for more
-	information.
+	information. To draw the progess indicator on top of the background
+	color, use the *over* attribute. To replace the background color, use
+	the *source* attribute (this can be useful when the notification is
+	semi-transparent).
 
-	Default: #5588AAFF
+	Default: over #5588AAFF
 
 *--markup* 0|1
 	If 1, enable Pango markup. If 0, disable Pango markup. If enabled, Pango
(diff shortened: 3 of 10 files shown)


main.c
```diff
@@ -28,6 +28,7 @@ static const char usage[] =
 	"                                  meaning as in CSS.\n"
 	"      --border-size <px>          Border size.\n"
 	"      --border-color <color>      Border color.\n"
+	"      --progress-color <color>    Progress bar color.\n"
```
vilhalmer: Should call it the "progress indicator" to be consistent with the long description in the manpage.

dbus/xdg.c
```diff
@@ -198,6 +198,10 @@ static int handle_notify(sd_bus_message *msg, void *data,
 			}
 			free(notif->desktop_entry);
 			notif->desktop_entry = strdup(desktop_entry);
+		} else if (strcmp(hint, "value") == 0) {
+			int32_t progress = 0;
+			ret = sd_bus_message_read(msg, "v", "i", &progress);
```
vilhalmer: Need an early return in here in case the read fails, like the other blocks have:
```
if (ret < 0) {
	return ret;
}
```

moverest:
I thought about drawing the progress indicator on top of the background. The reason I went for this option is that I use semi-transparent notifications and when you do that, you limit the range of color you can use for the progress element. Whereas, you can set whatever color you what when you divide it (you can still blend the color but you have to do it yourself). But, I think it could be useful to be able draw it on top as well as you said.

What do you think about an option `progress-mode` with the values: `divide`, `overlay` and `off`?

render.c
```diff
@@ -141,16 +141,36 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
...
-		offset_x + style->border_size,
+		offset_x + style->border_size + progress_width,
 		offset_y + style->border_size,
-		notif_width - border_size,
+		notif_background_width - progress_width,
 		notif_height - border_size,
 		scale);
 	cairo_fill(cairo);
 
+  // Render progress
```
emersion: Style: use tabs

render.c
```diff
@@ -250,7 +271,7 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 		format_text(style.format, text, format_state_text, state);
 
 		int hidden_height = render_notification(
-				cairo, state, &style, text, total_height, scale, NULL);
+                        cairo, state, &style, text, total_height, scale, NULL, 0);
```
emersion: Style: use tabs

render.c
```diff
@@ -212,7 +232,8 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 		}
 
 		int notif_height = render_notification(
-				cairo, state, style, text, total_height, scale, &notif->hotspot);
+					cairo, state, style, text, total_height, scale,
```
emersion: Style: indenting with one tab is enough for continuation lines

emersion:
>We do need a way to turn this off for people who don't want it, though.

Yeah. Your suggestion looks fine too.

>What do you think about an option `progress-mode` with the values: `divide`, `overlay` and `off`?

It would be easy to implement those using `cairo_save`, `CAIRO_OPERATOR_SOURCE`, and `cairo_restore`, without changing how the background is currently rendered.

Though, I'm not sure I like this option. Maybe it would be better as a global option, or as part of the color specification (`color = [blend|replace] #12345678`?). @vilhalmer Thoughts?

ammen99:
Hm, I am not familiar with how this works, but this looks a great idea. How can I test this?

moverest:
@ammen99 `notify-send Hello 'This is a notification.' --hint=int:value:34`.

vilhalmer:
> Though, I'm not sure I like this option. Maybe it would be better as a global option, or as part of the color specification (`color = [blend|replace] #12345678`?).

I like the color spec route. We could modify borders to draw over the background as well so that you can use it there. It also provides a neat new thing to play with in criteria.

It will need to modify how colors are stored, perhaps another layer of struct e.g.:
```
struct mako_color {
    uint32_t value;
    cairo_operator_t operator;
}
```
The operator should default to CAIRO_OPERATOR_OVER to maintain current behavior. We also don't need to provide keywords for [all the operator types](https://cairographics.org/manual/cairo-cairo-t.html#cairo-operator-t), we can add them as people want them. In the interest of keeping them extensible though we should probably keep their names somewhat close to cairo's (though the two described here don't collide with anything so I'm fine with those names as well).

moverest:
I'm ok with this. If you two are want that, I can push another commit with the changes.

emersion:
Yeah, I'm okay with this too. We can use Cairo's naming, and document that in the manpage.

I don't think it makes sense to use this struct with a color and an operator for all colors though. Could just use it for progress for now, and use it in a few more places later.

render.c
```diff
@@ -141,16 +141,36 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
...
 	cairo_fill(cairo);
 
+	// Render progress
+	set_source_u32(cairo, style->colors.progress);
+	set_rectangle(cairo,
+                offset_x + style->border_size,
+                offset_y + style->border_size,
+                progress_width,
+                notif_height - border_size,
+                scale);
```
ascent12: This whole indented part is done with spaces instead of tabs.

mako.1.scd
```diff
@@ -107,11 +107,13 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: #4C7899FF
 
-*--progress-color* _color_
+*--progress-color* [over|source] _color_
 	Set popup progress indicator color to _color_. See *COLOR* for more
-	information.
+	information. To draw the progess indicator on top of the background
+	color, use the *over* attribute. To replace the background color, use
+	the *source* attribute.
```
vilhalmer: Maybe note that `source` is mostly useful when the background is transparent, since it's not immediately obvious why the user should care about this distinction.

emersion:
Thanks!

PR #111, opened 2019-02-13
Grouping criteria

### LINKED ISSUE #66: Add option to group notifications with same title

vilhalmer:
Part 2 of 2! Closes #66. Also makes unicorns real.

Actual features:
- Renamed `group` to `group-by` and made it available on the cli.
- Grouping defaults to the new `none` criteria spec which never matches any notifications. This cannot be used in criteria sections because it would just be a fancy comment.
- Add an `invisible` style option to skip rendering arbitrary notifications. This is not exposed on the cli as it would make mako fairly useless. Maybe after we add history?
- Add %g format specifier for the group size.
- Inserts two new default criteria into the list:
    - Set the format of all grouped notifications to include the group size and make them invisible.
    - Re-visibilize the first one in the group.
- Groups count as a single notification when figuring out which ones to hide after `max-visible`.

Bonus feature:
- Store the raw criteria strings on criteria like sway does so they're available for easy reading when debugging.

Known issues:
- If multiple notifications per group are configured to be visible, only the first will be visible anyway when it is directly before the `max-visible` cutoff. I'm just going to file an issue for this since I don't expect many people to run into it.

Testing done:
- Lots, also probably not enough.
- Priority sorting still seems to work correctly, which I was worried about breaking.
- Did a bunch of reconfiguration stress testing to make sure things regrouped correctly and nothing exploded.

Things not done:
- I punted on converting `hidden` to a group. It's really hard to count the number of notifications outside of render. I still think this can be done, but it will take some more thought.
- I didn't end up adding a greater-than operator for criteria. I have it written, but it was getting complicated enough on its own that I left it out.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -302,9 +302,7 @@ bool apply_criteria_field(struct mako_criteria *criteria, char *token) {
 		return false;
 	}
 
-	// This should be unreachable.
-	fprintf(stderr, "Mysterious error parsing critiera field\n");
-	return false;
+	assert(false && "Criteria parser fell through");
 }
 
 // Retreive the global critiera from a given mako_config. This just so happens
--- a/criteria.c
+++ b/criteria.c
@@ -33,6 +33,7 @@ void destroy_criteria(struct mako_criteria *criteria) {
 	free(criteria->desktop_entry);
 	free(criteria->summary);
 	free(criteria->body);
+	free(criteria->raw_string);
 	free(criteria);
 }
 
@@ -189,6 +190,7 @@ bool parse_criteria(const char *string, struct mako_criteria *criteria) {
 		return false;
 	}
 
+	criteria->raw_string = strdup(string);
 	return true;
 }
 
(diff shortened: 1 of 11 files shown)


criteria.c
```diff
@@ -270,7 +302,9 @@ bool apply_criteria_field(struct mako_criteria *criteria, char *token) {
 		return false;
 	}
 
-	return true;
+	// This should be unreachable.
```
emersion: If unreachable, can be replaced with `assert(false)`

notification.c
```diff
@@ -379,7 +382,8 @@ int group_notifications(struct mako_state *state, struct mako_criteria *criteria
 	// is technically unnecessary, since it will go back in the same place, but
 	// it makes the rest of this logic nicer.
 	struct wl_list *location = NULL;  // The place we're going to reinsert them.
-	struct mako_notification *notif, *tmp;
+	struct mako_notification *notif = NULL, *tmp = NULL;
+	int count = 0;
```
emersion: This can be `size_t`

emersion:
Thanks!

PR #114, opened 2019-02-16
Return the same id when replacing a notification

### LINKED ISSUE #113: Replacing a notification shouldn't create a new ID
According to [Desktop Notifications Specification](https://developer.gnome.org/notification-spec/), the `replaces_id` parameter allows replacing active notifications.

> The optional notification ID that this notification replaces. The server must atomically (ie with no flicker or other visual cues) replace the given notification with this one. This allows clients to effectively modify the notification while it's active. A value of value of 0 means that this notification won't replace any existing notifications.

The following commands work on dunst and notify-osd, but do not work on mako:

```
$ gdbus call --session --dest org.freedesktop.Notifications --object-path /org/freedesktop/Notifications --method org.freedesktop.Notifications.Notify app_name 1 '' content1 '' '[]' '{}' 10000
$ gdbus call --session --dest org.freedesktop.Notifications --object-path /org/freedesktop/Notifications --method org.freedesktop.Notifications.Notify app_name 1 '' content2 '' '[]' '{}' 10000
```

vilhalmer:
This adds `reset_notification` and uses it to... reset the notification. Instead of closing and creating a new one, we put all of the new info into the old container so it retains its id and position in the list.

Fixes #113.

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -77,12 +77,6 @@ static void handle_notification_timer(void *data) {
 static int handle_notify(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct mako_state *state = data;
-
-	struct mako_notification *notif = create_notification(state);
-	if (notif == NULL) {
-		return -1;
-	}
-
 	int ret = 0;
 
 	const char *app_name, *app_icon, *summary, *body;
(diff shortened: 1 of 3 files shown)


dbus/xdg.c
```diff
@@ -93,23 +87,32 @@ static int handle_notify(sd_bus_message *msg, void *data,
...
+	struct mako_notification *notif = NULL;
+	if (replaces_id > 0) {
+		notif = get_notification(state, replaces_id);
+	}
+
+	if (notif) {
+		reset_notification(notif);
+	} else {
+		// Either we had no replaces_id, or the id given was invalid. Either
+		// way, make a new notification.
```
emersion: Maybe we can error out instead if the ID is invalid?

notification.c
```diff
@@ -25,6 +25,26 @@ bool hotspot_at(struct mako_hotspot *hotspot, int32_t x, int32_t y) {
 		y < hotspot->y + hotspot->height;
 }
 
+void reset_notification(struct mako_notification *notif) {
```
emersion: Hmm, this doesn't really reset, it frees everything inside the notification. To really reset it, would it be possible to set free'd pointers to NULL to make sure we don't use-after-free?

dbus/xdg.c
vilhalmer: The spec doesn't have anything to say about this, so I thought it might be better to err on the side of not punishing a client that misses replacing it by just a moment. For example, the `notify-send` replacement that I use tracks its replacement id in a file, which is not super robust but works for a client that doesn't run continuously.

notification.c
vilhalmer: Yep, good point.

dbus/xdg.c
emersion: That's right, I agree.

emersion:
Thanks!

Issue #37, opened 2018-05-23
Support icons


smlx:
It would be great to have icon support in mako. What do you think?

emersion:
See https://developer.gnome.org/notification-spec/#icons-and-images

I'm not against it, but I also don't want to add huge piles of code to support it. I personally don't care about it, so patches welcome!

Issue #91, opened 2018-11-05
History and history redisplay


Emantor:
Feature request:
Keep a history of configurable length which can be redisplayed using makoctl (similar to dunsts history handling)

vilhalmer:
I am a fan of having them accessible via makoctl, though I'd probably lean toward having both methods available and having a `makoctl redisplay N` or similar for the dunst behavior instead of a built-in keybinding. Max history length of -1 could mean "infinite" as well.

Difrex:
Any progress on it? It's a really important feature that I need.

emersion:
Feel free to send patches.

MichaelAquilina:
So I've started giving the source code a look to see if I could add this in myself. Hooking into `dbus/mako.c` looks simple enough. However it seems like we currently do not track history of notifications in any way - so that would have to change. Is there any way you would suggest looking into this?

vilhalmer:
I haven't put a whole lot of thought into this, but the simplest implementation would probably be another linked list on the mako struct where we put notifications instead of destroying them. I don't think you need to worry about ever updating their styles, since they're a "thing of the past" so to speak. Then when you want to redisplay them, you could simply grab them in reverse order and put them back into the normal list.

One of the concerns I do have is using a large chunk of memory when icons/images are present, but I don't think we need to tackle that for the first attempt. Better to see what problems really arise after using it for a while.

matt-snider:
@MichaelAquilina @vilhalmer Consider having a look at the implementation used in [dunst](https://github.com/dunst-project/dunst).

From a quick look over the code these seem to be some of the important bits:

* Notification is created [here](https://github.com/dunst-project/dunst/blob/59a87e4e998d89c6571ff0558aa261d247c7636c/src/dbus.c#L315)
    * Calls [queues_notification_insert()](https://github.com/dunst-project/dunst/blob/6c4eeda43412f0ecd0e38e00069ca47aff4452a8/src/queues.c#L160) which adds it to the pending notification list
* Notification is dismissed [here](https://github.com/dunst-project/dunst/blob/ac94eca79e37a50c7b531f2706de631c24f86276/src/x11/x.c#L307) when key bound to dismiss is pressed
    * Calls [queues_notification_close()](https://github.com/dunst-project/dunst/blob/6c4eeda43412f0ecd0e38e00069ca47aff4452a8/src/queues.c#L334)
[...]

PR #115, opened 2019-02-16
Partial icon support

algmyr:
image

Adds preliminary support for icons, assuming they are given as an image path. Support for the other options in https://developer.gnome.org/notification-spec/#icons-and-images could be worked on later.

Image loading code is mostly borrowed from swaybg and adds a dependency on gdk-pixbuf. This could be made an optional dependency, I removed that aspect for the sake of simplicity when writing the code.

Some changes to the present rendering code had to be done
* If no icon is present (or if it failed to load) then the behavior would be unchanged
* If an icon is present a spacing of padding.left is given on both sides (padding, icon, padding, text)
* The icon and text are both vertically centered, so it looks nice if text is taller than the icon and vice versa

Additional icon logic
* The icon is fitted (maintaining aspect ratio) inside a square of `max_icon_size`

Adding support for the other icon formats in the spec wouldn't change the rendering logic, so hopefully this can be merged independently. Some code would have to change to pass relevant parameters, but that's about it.

Partially resolves https://github.com/emersion/mako/issues/37.

### DIFF
--- a/config.c
+++ b/config.c
@@ -94,6 +94,8 @@ void init_default_style(struct mako_style *style) {
 
 	style->border_size = 2;
 
+	style->max_icon_size = 64;
+
 	style->font = strdup("monospace 10");
 	style->markup = true;
 	style->format = strdup("<b>%s</b>\n%b");
@@ -176,6 +178,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 		target->spec.border_size = true;
 	}
 
+	if (style->spec.max_icon_size) {
+		target->max_icon_size = style->max_icon_size;
+		target->spec.max_icon_size = true;
+	}
+
 	if (style->spec.font) {
 		free(target->font);
 		target->font = new_font;
@@ -254,6 +261,7 @@ bool apply_superset_style(
 	target->spec.margin = true;
 	target->spec.padding = true;
 	target->spec.border_size = true;
+	target->spec.max_icon_size = true;
 	target->spec.default_timeout = true;
 	target->spec.markup = true;
 	target->spec.actions = true;
@@ -287,6 +295,7 @@ bool apply_superset_style(
 			max(style->padding.bottom, target->padding.bottom);
 		target->padding.left = max(style->padding.left, target->padding.left);
 		target->border_size = max(style->border_size, target->border_size);
+		target->max_icon_size = max(style->max_icon_size, target->max_icon_size);
 		target->default_timeout =
 			max(style->default_timeout, target->default_timeout);
 
@@ -412,6 +421,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 		return spec->colors.border = parse_color(value, &style->colors.border);
 	} else if (strcmp(name, "progress-color") == 0) {
 		return spec->colors.progress = parse_mako_color(value, &style->colors.progress);
+	} else if (strcmp(name, "max-icon-size") == 0) {
+		return spec->max_icon_size =
+			parse_int(value, &style->max_icon_size);
 	} else if (strcmp(name, "markup") == 0) {
 		return spec->markup = parse_boolean(value, &style->markup);
 	} else if (strcmp(name, "actions") == 0) {
@@ -579,6 +591,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"border-size", required_argument, 0, 0},
 		{"border-color", required_argument, 0, 0},
 		{"progress-color", required_argument, 0, 0},
+		{"max-icon-size", required_argument, 0, 0},
 		{"markup", required_argument, 0, 0},
 		{"actions", required_argument, 0, 0},
 		{"format", required_argument, 0, 0},
--- a/config.c
+++ b/config.c
@@ -94,8 +94,13 @@ void init_default_style(struct mako_style *style) {
 
 	style->border_size = 2;
 
+#ifdef HAVE_ICONS
 	style->icons = true;
+#else
+	style->icons = false;
+#endif
 	style->max_icon_size = 64;
+
 
 	style->font = strdup("monospace 10");
 	style->markup = true;
@@ -302,7 +307,7 @@ bool apply_superset_style(
 			max(style->padding.bottom, target->padding.bottom);
 		target->padding.left = max(style->padding.left, target->padding.left);
 		target->border_size = max(style->border_size, target->border_size);
-		target->icons = style->max_icon_size || target->max_icon_size;
+		target->icons = style->icons || target->icons;
 		target->max_icon_size = max(style->max_icon_size, target->max_icon_size);
 		target->default_timeout =
 			max(style->default_timeout, target->default_timeout);
@@ -430,8 +435,13 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	} else if (strcmp(name, "progress-color") == 0) {
 		return spec->colors.progress = parse_mako_color(value, &style->colors.progress);
 	} else if (strcmp(name, "icons") == 0) {
+#ifdef HAVE_ICONS
 		return spec->icons =
 			parse_boolean(value, &style->icons);
+#else
+		fprintf(stderr, "Icon support not built in, ignoring icons setting.\n");
+		return true;
+#endif
 	} else if (strcmp(name, "max-icon-size") == 0) {
 		return spec->max_icon_size =
 			parse_int(value, &style->max_icon_size);
--- a/config.c
+++ b/config.c
@@ -94,6 +94,7 @@ void init_default_style(struct mako_style *style) {
 
 	style->border_size = 2;
 
+	style->show_icons = true;
 	style->max_icon_size = 64;
 
 	style->font = strdup("monospace 10");
@@ -176,6 +177,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	if (style->spec.border_size) {
 		target->border_size = style->border_size;
 		target->spec.border_size = true;
+	}
+
+	if (style->spec.show_icons) {
+		target->show_icons = style->show_icons;
+		target->spec.show_icons = true;
 	}
 
 	if (style->spec.max_icon_size) {
@@ -261,6 +267,7 @@ bool apply_superset_style(
 	target->spec.margin = true;
 	target->spec.padding = true;
 	target->spec.border_size = true;
+	target->spec.show_icons = true;
 	target->spec.max_icon_size = true;
 	target->spec.default_timeout = true;
 	target->spec.markup = true;
@@ -295,6 +302,7 @@ bool apply_superset_style(
 			max(style->padding.bottom, target->padding.bottom);
 		target->padding.left = max(style->padding.left, target->padding.left);
 		target->border_size = max(style->border_size, target->border_size);
+		target->show_icons = style->max_icon_size || target->max_icon_size;
 		target->max_icon_size = max(style->max_icon_size, target->max_icon_size);
 		target->default_timeout =
 			max(style->default_timeout, target->default_timeout);
@@ -421,6 +429,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 		return spec->colors.border = parse_color(value, &style->colors.border);
 	} else if (strcmp(name, "progress-color") == 0) {
 		return spec->colors.progress = parse_mako_color(value, &style->colors.progress);
+	} else if (strcmp(name, "show-icons") == 0) {
+		return spec->show_icons =
+			parse_boolean(value, &style->show_icons);
 	} else if (strcmp(name, "max-icon-size") == 0) {
 		return spec->max_icon_size =
 			parse_int(value, &style->max_icon_size);
@@ -591,6 +602,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"border-size", required_argument, 0, 0},
 		{"border-color", required_argument, 0, 0},
 		{"progress-color", required_argument, 0, 0},
+		{"show-icons", required_argument, 0, 0},
 		{"max-icon-size", required_argument, 0, 0},
 		{"markup", required_argument, 0, 0},
 		{"actions", required_argument, 0, 0},
--- a/config.c
+++ b/config.c
@@ -94,7 +94,7 @@ void init_default_style(struct mako_style *style) {
 
 	style->border_size = 2;
 
-	style->show_icons = true;
+	style->icons = true;
 	style->max_icon_size = 64;
 
 	style->font = strdup("monospace 10");
@@ -179,9 +179,9 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 		target->spec.border_size = true;
 	}
 
-	if (style->spec.show_icons) {
-		target->show_icons = style->show_icons;
-		target->spec.show_icons = true;
+	if (style->spec.icons) {
+		target->icons = style->icons;
+		target->spec.icons = true;
 	}
 
 	if (style->spec.max_icon_size) {
@@ -267,7 +267,7 @@ bool apply_superset_style(
 	target->spec.margin = true;
 	target->spec.padding = true;
 	target->spec.border_size = true;
-	target->spec.show_icons = true;
+	target->spec.icons = true;
 	target->spec.max_icon_size = true;
 	target->spec.default_timeout = true;
 	target->spec.markup = true;
@@ -302,7 +302,7 @@ bool apply_superset_style(
 			max(style->padding.bottom, target->padding.bottom);
 		target->padding.left = max(style->padding.left, target->padding.left);
 		target->border_size = max(style->border_size, target->border_size);
-		target->show_icons = style->max_icon_size || target->max_icon_size;
+		target->icons = style->max_icon_size || target->max_icon_size;
 		target->max_icon_size = max(style->max_icon_size, target->max_icon_size);
 		target->default_timeout =
 			max(style->default_timeout, target->default_timeout);
@@ -429,9 +429,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 		return spec->colors.border = parse_color(value, &style->colors.border);
 	} else if (strcmp(name, "progress-color") == 0) {
 		return spec->colors.progress = parse_mako_color(value, &style->colors.progress);
-	} else if (strcmp(name, "show-icons") == 0) {
-		return spec->show_icons =
-			parse_boolean(value, &style->show_icons);
+	} else if (strcmp(name, "icons") == 0) {
+		return spec->icons =
+			parse_boolean(value, &style->icons);
 	} else if (strcmp(name, "max-icon-size") == 0) {
 		return spec->max_icon_size =
 			parse_int(value, &style->max_icon_size);
@@ -602,7 +602,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"border-size", required_argument, 0, 0},
 		{"border-color", required_argument, 0, 0},
 		{"progress-color", required_argument, 0, 0},
-		{"show-icons", required_argument, 0, 0},
+		{"icons", required_argument, 0, 0},
 		{"max-icon-size", required_argument, 0, 0},
 		{"markup", required_argument, 0, 0},
 		{"actions", required_argument, 0, 0},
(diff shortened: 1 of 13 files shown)


algmyr:
That CI fail is from the missing gdk-pixbuf, as expected.

emersion:
Thanks for your pull request! Can you rebase it, I've just merged another one and there is a conflict?

Also, before this is merged, I'd prefer to have a Meson option to disable `gdk-pixbuf`.

icon.c
```diff
@@ -0,0 +1,161 @@
+#include <stdio.h>
+#include <assert.h>
+#include <cairo/cairo.h>
+#include <gdk-pixbuf/gdk-pixbuf.h>
+
+#include "icon.h"
+
+cairo_surface_t* gdk_cairo_image_surface_create_from_pixbuf(const GdkPixbuf *gdkbuf) {
```
emersion: Hmm, I'm not a fan of this function.

algmyr: Name-wise or functionalify-wise? The name is unchanged from swaybg, and the function itself is the same apart from logging changes and some ifdefs.

algmyr: My bad. Apparently I based the name off of a PR and not the current code. I could make it more similar to https://github.com/swaywm/sway/blob/master/common/background-image.c if that helps?

vilhalmer: Is this different from `gdk_cairo_surface_create_from_pixbuf`? https://developer.gnome.org/gdk3/stable/gdk3-Cairo-Interaction.html#gdk-cairo-surface-create-from-pixbuf`

emersion: Yeah, using `gdk_cairo_image_surface_create_from_pixbuf` would be a lot better :)

vilhalmer: There's also `gdk_cairo_set_source_pixbuf` which I use in oguri. I'm not familiar enough to know what the particular upsides and downsides of each method are.

vilhalmer:
(I'll hopefully be able to review this later today, at the very least sometime this weekend.)

icon.c
algmyr: I guess the downside of both those is a dependence also on gdk (although that can be made optional along with gdk-pixbuf). For the latter I assume that would be used in `render.c`? If so it might be nicer to do the former (assuming it works properly) and return a cairo surface as I do now. That way all the ifdefs related to optional gdk stuff is constrained to `icon.c`.

algmyr:
That should be the rebase done. Code rewritten to use `gdk_cairo_set_source_pixbuf`. All ifdefs are contained in `icon.c` and `icon.h`. Flag added to compile with/without icons, defaults to compile with icons if the dependencies are there. It's worth noting that using the gdk function drags in gtk3 as a dependency as well (same as in oguri). I guess that's why sway and other project decided to roll their own converter instead of having gtk3 as a dependency.

emersion:
Instead of having guards in `icon.{c,h}`, is it possible to move them in `render.c`?

emersion:
Would be nice to have an option to disable completely icons too.

algmyr:
> Would be nice to have an option to disable completely icons too.

As a boolean provided at runtime?

algmyr:
Ok, guards moved and runtime option added.

Also, when noticing I had missed adding an option to the help I also saw `--ignore-timeout` is not listed in there. Not sure if it's intentional or not.

render.c
```diff
@@ -131,7 +147,16 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	}
 	int text_height = buffer_text_height / scale;
 
-	int notif_height = border_size + padding_height + text_height;
+	int notif_height;
+#ifdef SHOW_ICONS
```
vilhalmer: Does this really need a guard? If icons are compiled out, the image would just always be NULL.

render.c
```diff
@@ -178,12 +203,24 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	cairo_fill(cairo);
 	cairo_restore(cairo);
 
+#ifdef SHOW_ICONS
```
vilhalmer: Same question re: guard.

render.c
```diff
@@ -88,11 +93,22 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 		offset_x = style->margin.left;
 	}
 
+	double text_x = style->padding.left;
+#ifdef SHOW_ICONS
+	struct mako_icon icon;
+	if (style->show_icons) {
+		icon = get_icon(icon_path, style->max_icon_size);
```
vilhalmer: This should probably be done in `handle_notify`, rather than pulling the icon from disk every frame. The loaded image would then get cached in the `mako_notification`. This will end up reading the image even if icons are disabled, but I think this is a fair trade to avoid needing to scatter logic around other places to check if it gets changed by criteria and load the image then.

icon.c
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return icon;
+}
+
+void draw_icon(cairo_t *cairo, struct mako_icon icon,
+		double xpos, double ypos, double scale) {
+	cairo_save(cairo);
+	cairo_scale(cairo, icon.scale * scale, icon.scale * scale);
+	gdk_cairo_set_source_pixbuf(cairo, icon.image, xpos / icon.scale, ypos / icon.scale);
```
vilhalmer: I wonder if it would be more efficient to create a new cairo surface and draw the icon into it when loading it, rather than waiting until we draw it to convert from the pixbuf? I don't have strong feelings one way or the other about this though.

include/icon.h
```diff
@@ -0,0 +1,22 @@
...
+#include <cairo/cairo.h>
+#include <gdk-pixbuf/gdk-pixbuf.h>
+
+struct mako_icon {
+	double width;
+	double height;
+	double image_width;
+	double image_height;
+	double scale;
+	GdkPixbuf* image;
```
vilhalmer: Style: asterisk should be on the other side of the space.

icon.c
```diff
@@ -0,0 +1,69 @@
...
+	cairo_paint(cairo);
+	cairo_restore(cairo);
+}
+
+void destroy_icon(struct mako_icon icon) {
+	if (icon.image !=  NULL) {
```
vilhalmer: Style: extra space after the operator.

icon.c
```diff
@@ -0,0 +1,69 @@
...
+		fprintf(stderr, "Failed to load icon (%s)\n", err->message);
+		return NULL;
+	}
+	return pixbuf;
+}
+
+struct mako_icon get_icon(const char *path, double max_size) {
+	struct mako_icon icon;
+	icon.image = load_image(path);
+	if (icon.image == NULL)
```
vilhalmer: Style: please use braces even for one-line conditionals.

vilhalmer:
> Also, when noticing I had missed adding an option to the help I also saw `--ignore-timeout` is not listed in there. Not sure if it's intentional or not.

This is not intentional.

meson.build
```diff
@@ -33,27 +33,42 @@ else
 	add_project_arguments('-DHAVE_ELOGIND=1', language : 'c')
 endif
 
+gdk = dependency('gdk-3.0', required: get_option('icon'))
+gdk_pixbuf = dependency('gdk-pixbuf-2.0', required: get_option('icon'))
+icon_feature = not get_option('icon').disabled() and gdk.found() and gdk_pixbuf.found()
```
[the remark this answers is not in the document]
algmyr: Also in the case of default case of `icon_feature='auto'`?

algmyr:
Is `SHOW_ICONS` a good name for the define? I just named is whatever at the time. Maybe something like `ICON_SUPPORT` would be nicer?

render.c
algmyr: As it's written now when icons are disabled every mention of them disappear.

icon.c
algmyr: It might be a good idea to not only make it into a cairo surface, but also to scale it immediately as well (especially if we move loading to `handle_notify`). **Pro:** doesn't store the full resolution icon for long. **Con:** in the odd case of the user changing dpi or `max-icon-size` at runtime an active notification might have a too low res icon.

render.c
algmyr: This is also good preparation for adding support for icons via `icon_data` later on.

algmyr:
A bit conflicted about where to do the ifdefs. Original thought was to put them in `icon.{c,h}` and let a neutered icon struct be left in the code. After a suggestion I moved them to render.c.

Now, moving things to `handle_notify` and into the notifications that means quite a few more ifdefs. It's probably better to go back to a neutered icon struct.

algmyr:
Made some larger changes when moving creation and destruction into notification code and such. 

Quick overview:
* a `mako_icon` is now part of the notification as a pointer
* functions in `icon.c` now operate on icon pointers
* all ifdefs in the code are gone except for one in `icon.c`
* the icons images are loaded, painted (at the correct size) to a cairo surface, and then released

I think (hope) this cleaned up a lot of code (fewer ifdefs, better structure in `icon.c`). But this essentially means looking over most of the code another round to check that I didn't introduce anything bad again. Sorry about that.

vilhalmer:
> But this essentially means looking over most of the code another round to check that I didn't introduce anything bad again. Sorry about that.

On the contrary, thanks for working through the change requests. :)

> Is output scale known when `handle_notify` happens? If so I can fix that.

Hmm, this is a good point that I hadn't considered. The output scale is only known if we know which output we're displaying on, which in the default mode can change whenever there are zero notifications open and thus no surface. Perhaps it should scale to the largest factor of any of the outputs connected when loading the image, then scale down in `render` if the chosen output is smaller than this? (Does this happen automatically if you draw the cairo surface into a smaller target area?) That still prevents needing to scale from the full source size each time, and in the happy case only does it once. Sorry to give you the run-around on this, but I'm glad you caught it.

This still isn't perfect, as if another display is connected it could have a larger scale factor. @emersion any thoughts on how clever we want to be here?

render.c
```diff
@@ -339,7 +321,7 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 		format_text(style.format, text, format_hidden_text, &data);
 
 		int hidden_height = render_notification(
-			cairo, state, &style, text, "", total_height, scale, NULL, 0);
+			cairo, state, &style, text, notif->icon, total_height, scale, NULL, 0);
```
vilhalmer: This should be NULL since there's no valid notification attached to this fake one.

algmyr:
>Perhaps it should scale to the largest factor of any of the outputs connected when loading the image, then scale down in `render` if the chosen output is smaller than this?

That's pretty much what I've considered. I'll have to make the `draw_icon` code a bit smarter since it can no longer assume what it gets has the final resolution. Maybe the icon struct could have a member `assumed_output_scale` or something. If so

    cairo_scale(cairo, scale, scale);

would be changed into something like

    double sc = scale/icon->assumed_output_scale;
    cairo_scale(cairo, sc, sc);

Current code is equivalent to `assumed_output_scale = 1`.

>(Does this happen automatically if you draw the cairo surface into a smaller target area?)
Not really sure what you mean there. The scale I set in `draw_icon` is the scale it will be drawn at. If it's larger than the surface then the image will be clipped.

>That still prevents needing to scale from the full source size each time, and in the happy case only does it once.
I don't think scaling in cairo should affect performance much (but don't quote me on that). Isn't it just a issue of sampling the source surface when painting? We already paint it from the source surface all the time in `draw_icon`.

The biggest advantage of resizing beforehand would be memory usage, as the full size image doesn't have to be kept in the cairo surface in the icon struct. I may be guilty of sending of sending full sized image artwork (~1500x1500) for the notification deamon to handle.

Resizing immediately would mean a short spike in memory when reading the image that goes away immediately after painting the icon sized image to the cairo surface. Not resizing immediately means keeping that full sized image in memory until the notification has disappeared.

algmyr:
We could also go back to storing the full sized image and taking the hit to memory usage. This removes all worries about making the icon too small.

This would essentially be letting the user take the consequence for sending in that 100MiB png file as an icon. We would have had to parse it to a pixbuf anyway, so we'll not be able to get rid of the initial memory spike.

I'm quite torn about how to deal with it. Keeping it full sized guarantees we'll be able to do the right thing, scaling it down based on maximum output scale increases the code complexity a bit and there is always the chance that we get things wrong in some odd special cases.

An odd special case is having a notification on screen and changing the output scale. The current code is smart enough to scale things correctly, but will end up with a blurred mess of an icon.

vilhalmer:
> We could also go back to storing the full sized image and taking the hit to memory usage. This removes all worries about making the icon too small.
> 
> This would essentially be letting the user take the consequence for sending in that 100MiB png file as an icon. We would have had to parse it to a pixbuf anyway, so we'll not be able to get rid of the initial memory spike.

My main concern with this method is the inevitability of a history feature (#91). Keeping a huge number of original-size images hanging around is a recipe for disaster. For disk-sourced icons we can keep them as a path instead and reload them if they are redisplayed, but we won't be able to do that if they're sent as data. On the other hand, we can always deal with that when it comes around.

> I'm quite torn about how to deal with it. Keeping it full sized guarantees we'll be able to do the right thing, scaling it down based on maximum output scale increases the code complexity a bit and there is always the chance that we get things wrong in some odd special cases.

I agree, let's do the easy thing for now and if it becomes a problem we can revisit it then.

> An odd special case is having a notification on screen and changing the output scale. The current code is smart enough to scale things correctly, but will end up with a blurred mess of an icon.

Changing output scale is a weird enough case that I don't think it's worth worrying about.

algmyr:
I think that should cover it. I've tried running with a different output scale and the notifications look fine. 

Now, a remark that is not related to this PR:
I think the border rendering is a bit off. This is an issue I kinda felt was there from the beginning (the inside of the border sometimes look see through) and the border didn't look like it aligned with the rest of the notification (the border seems left shifted slightly). Looking at the upscaled version it's a lot more apparent (excuse the fonts, pixel fonts doesn't scale well with antialiasing):
image
This is not an issue with this PR, it's there in master as well.

meson.build
```diff
@@ -33,27 +33,39 @@ else
 	add_project_arguments('-DHAVE_ELOGIND=1', language : 'c')
 endif
 
+gdk = dependency('gdk-3.0', required: get_option('icon'))
+gdk_pixbuf = dependency('gdk-pixbuf-2.0', required: get_option('icon'))
+icon_feature = not get_option('icon').disabled()
+if icon_feature
```
emersion: This won't work if `-Dicon=auto` and gdk-pixbuf isn't found. Instead, the `if` should be changed to:

```meson
if gdk.found() and gdk_pixbuf.found()
```

meson.build
```diff
@@ -33,27 +33,39 @@ else
...
+	add_global_arguments('-DSHOW_ICONS=1', language : 'c')
```
emersion: I'd prefer to rename this to `HAVE_GDK_PIXBUF`

meson_options.txt
```diff
@@ -0,0 +1 @@
+option('icon', type: 'feature', value: 'enabled', description: 'Enable icon support')
```
emersion: This should default to `auto`.

main.c
```diff
@@ -29,6 +29,8 @@ static const char usage[] =
 	"      --border-size <px>          Border size.\n"
 	"      --border-color <color>      Border color.\n"
 	"      --progress-color <color>    Progress indicator color.\n"
+	"      --icons <0|1>          Show icons in notifications.\n"
```
emersion: Indentation is a little bit off

icon.c
```diff
@@ -0,0 +1,76 @@
...
+#include <assert.h>
+#include <cairo/cairo.h>
+
+#include "icon.h"
+
+#ifdef SHOW_ICONS
+#include <gdk/gdk.h>
+#include <gdk-pixbuf/gdk-pixbuf.h>
+
+GdkPixbuf *load_image(const char *path) {
```
emersion: A few functions here can be `static`

icon.c
```diff
@@ -0,0 +1,76 @@
...
+	if (strlen(path) == 0) {
+		return NULL;
+	}
+	GError *err = NULL;
+	GdkPixbuf *pixbuf = gdk_pixbuf_new_from_file(path, &err);
+	if (!pixbuf) {
+		fprintf(stderr, "Failed to load icon (%s)\n", err->message);
```
emersion: Need to free the error in this case

dbus/xdg.c
```diff
@@ -109,6 +111,7 @@ static int handle_notify(sd_bus_message *msg, void *data,
 	notif->app_icon = strdup(app_icon);
 	notif->summary = strdup(summary);
 	notif->body = strdup(body);
+	notif->icon = create_icon(notif->app_icon, state->config.superstyle.max_icon_size);
```
emersion: Is it possible to check for `style->icons` here instead of in the rendering functions? This would allow to reduce memory usage in case icons are disabled. The downside is that dynamically changing the configuration to enable icons won't draw icons for existing notifications, but I think that's fine.

meson.build
algmyr: This has been discussed a bit back and forth (see https://github.com/emersion/mako/pull/115#discussion_r257512600). What is the wanted behavior? As soon as I know that for sure I can implement it. Is it:

    has dep | -Dicon   | outcome
    ----------------------------------
    yes     | enabled  | with icons
    no      | enabled  | error?
    yes     | auto     | with icons
    no      | auto     | without icons
    yes     | disabled | without icons
    no      | disabled | without icons

emersion: Meson features will make the build fail if `-Dicons=enabled` and GDK isn't available. This is what `required: get_option('icon')` does. More info: http://mesonbuild.com/Build-options.html#features

emersion: By the way, nitpick: I'd rename the `icon` option to `icons`, for consistency.

emersion:
Oh, if `style->icons` is true, we should probably advertise the `icon-static` capability.

meson.build
algmyr: Is it clear that this relates to having both gdk and gdk_pixbuf? They are largely independent things.

emersion: Okay, maybe `HAVE_ICONS` then?

algmyr: Sounds good to me.

meson.build
algmyr: Got it. Haven't worked with meson much outside some trivial projects, so it's a learning experience.

dbus/xdg.c
algmyr: Yes, that should work. It makes the checks in `render.c` nicer too. Good catch.

algmyr:
> Oh, if `style->icons` is true, we should probably advertise the `icon-static` capability.

What about the case where we have don't have support for icons, but when `style->icons` is still true?

emersion:
>Is adding an ifdef around the capability part a reasonable solution?

Maybe we should disallow setting `style->icons` to true when built without icons support?

algmyr:
> Maybe we should disallow setting `style->icons` to true when built without icons support?

So, in the case of no icon support: Make the default value of `style->icons` false and then remove all other places where it can be set?

emersion:
>Make the default value of `style->icons` false and then remove all other places where it can be set?

Yeah, but I think there's only one such place. Also, we could `assert(false)` in `create_icon`.

algmyr:
Something like this?

    #ifdef HAVE_ICONS
		return spec->icons =
			parse_boolean(value, &style->icons);
    #else
		bool b;
		bool success = parse_boolean(value, &b);
		if (b) {
			fprintf(stderr, "Icon support not built in, ignoring icons setting");
		}
		return success;
    #endif

Feels a bit clumsy.

config.c
```diff
@@ -287,6 +302,8 @@ bool apply_superset_style(
 			max(style->padding.bottom, target->padding.bottom);
 		target->padding.left = max(style->padding.left, target->padding.left);
 		target->border_size = max(style->border_size, target->border_size);
+		target->icons = style->max_icon_size || target->max_icon_size;
```
algmyr: I'm impressed no one noticed this gem of a copy-paste error. Oh well, better late than never.

algmyr:
I commited the changes I felt were reasonable. Will amend if needed.

config.c
emersion: Eh, good catch

emersion:
Thanks!

algmyr:
Thanks to you two as well! It took some time, but the merged feature is vastly improved compared to what it started out as.

smlx:
This is great, thank you everyone!

lhanson:
Thanks from a happy user!

vilhalmer:
Thanks for your work on this @algmyr!

> Oh, if `style->icons` is true, we should probably advertise the `icon-static` capability.

Good catch, I totally forgot about capabilities.

PR #87, opened 2018-10-25
Add center position for notifications

hpdeifel:
There you go, fixes #86 

I didn't add vertical centers for now. I don't think they'd be particularly useful, if anyone cares they can always add them later.

PR #120, opened 2019-03-07
Add true 'center' anchor option, and re-organize order of options

StephenBrown2:
Inspired by https://github.com/emersion/mako/pull/87, and the fact that I use a 39" 4k TV as my monitor, I've added a true 'center' option to the anchor positions.

I also noticed that the order of options wasn't grouped as I thought it should be, so I fixed that as well. Now it goes Right, Center, Left, for both Top and Bottom. Happy to back out that change though if you feel it should be separate. ~I plan on making a PR to change the check to a switch statement next.~ Strike that, I didn't know C couldn't switch compare strings.


emersion:
Thanks!

Issue #118, opened 2019-02-20
Notification images support


emersion:
Add support for "image-path" and "image-data". gdk-pixbuf can probably do most of the work.

> The "image-data" and "icon_data" hints should be a raw image data structure of signature (iiibiiay) which describes the width, height, rowstride, has alpha, bits per sample, channels and image data respectively.
>
>The "app_icon" parameter and "image-path" hint should be either an URI (file:// is the only URI schema supported right now) or a name in a freedesktop.org-compliant icon theme (not a GTK+ stock ID).

emersion:
"app_icon" support is incomplete:

```
Failed to load icon (Failed to open file “weechat”: No such file or directory)
```

emersion:
I think what's left is support for "image-data" and "icon_data".

emersion:
https://github.com/emersion/mako/pull/147 added support for "image-data". Since "icon_data" is deprecated, let's just close this issue.

PR #127, opened 2019-03-22
Add support for loading images by icon-name

sre:
See issue #118 . This adds a dependency to gtk+-3.0, since the icon features are unfortunately not available in gdk. The code tries to load the argument as file/path first and otherwise falls back to icon-name based lookup otherwise.

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -8,6 +8,7 @@
 #ifdef HAVE_ICONS
 #include <gdk/gdk.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
+#include <gtk/gtk.h>
 
 static GdkPixbuf *load_image(const char *path) {
 	if (strlen(path) == 0) {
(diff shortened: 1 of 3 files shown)


vilhalmer:
Hi! Thanks for this, it looks like a pretty clean implementation. @emersion and I have been talking out of band about whether we want to actually pull in GTK/GDK as a dependency after the work to add icons, or if we want to pull it back out before we ship 1.3. It was only added to avoid a single function that manually rearranges some pixels. If we decide to pull it back out, we'd unfortunately need to implement this by hand as well. I've been working on it on and off and have a mostly functional branch locally.

I won't close this until we reach a decision, since if we do keep GTK this is obviously much cleaner than doing it ourselves. Sorry for the bait-and-switch, I should have assigned myself to #118.

emersion:
Thanks for your pull request! Unfortunately work has been duplicated and https://github.com/emersion/mako/pull/128 has been merged. Really sorry about that! :S

PR #128, opened 2019-03-24
Implement resolve_icon to support icon names

vilhalmer:
Alternative implementation of #127 to fix part of #118, in case we decide to ditch GTK. This also fixes disabling icons within criteria.

I haven't actually gotten to test scaled icons, I don't have a hidpi display available at the moment or even any scaled icons to display thereon. I did make a mockup of an icon theme to verify that they were selected correctly if I forced the scale to 2x, though.


dbus/xdg.c
```diff
@@ -283,6 +280,10 @@ static int handle_notify(sd_bus_message *msg, void *data,
 			handle_notification_timer, notif);
 	}
 
+	if (notif->style.icons) {
+		notif->icon = create_icon(notif);
+	}
+
```
vilhalmer: This will prevent icons from appearing on existing notifications after a reload if they are re-enabled, but I think that's a fair trade for never trying to hit disk if a user has them disabled.

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+//   in the icon theme spec (https://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html)
+//
+// Returns the resolved path, or NULL if it was unable to find an icon. The
+// return value must be freed by the caller.
+static char *resolve_icon(struct mako_notification *notif) {
+	char *icon_name = notif->app_icon;
+	if (icon_name[0] == '\0') {
+		return NULL;
+	}
+	if (icon_name[0] == '/' || icon_name[0] == '~') {
```
emersion: I don't think `~` will work. Probably better to leave it out for now.

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+		return strdup(icon_name);
+	}
+	if (strstr(icon_name, "file://") == icon_name) {
+		// Just chop off the scheme.
+		return strdup(icon_name + 7);
```
emersion: Is it possible to use `strlen("file://")` instead of this magic number?

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	// Determine the largest scale factor of any attached output.
+	int32_t max_scale = 1;
+	struct mako_output *output = NULL;
+	wl_list_for_each(output, &notif->state->outputs, link) {
+		if (output->scale > max_scale) {
+			max_scale = output->scale;
+		}
+	}
+
+	static const char *fallback = "%s:/usr/share/icons/hicolor";
```
emersion: This can be an array (`static const char fallback[] = …`)

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	char *theme_path = strtok_r(search, ":", &saveptr);
+
+	// Match all icon files underneath of the theme_path followed by any icon
+	// size and category subdirectories. This pattern assumes that all the
+	// files in the icon path are valid icon types.
+	static const char *pattern_fmt = "%s/*/*/%s.*";
+
+	char *icon_path = NULL;
+	int32_t last_icon_size = 0;
+	while (theme_path) {
```
emersion: We also probably want to skip empty `theme_path`s (this happens if `notif->style.icon_path` is empty for instance)

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	static const char *fallback = "%s:/usr/share/icons/hicolor";
+	char *search = mako_asprintf(fallback, notif->style.icon_path);
+
+	char *saveptr = NULL;
+	char *theme_path = strtok_r(search, ":", &saveptr);
+
+	// Match all icon files underneath of the theme_path followed by any icon
+	// size and category subdirectories. This pattern assumes that all the
+	// files in the icon path are valid icon types.
+	static const char *pattern_fmt = "%s/*/*/%s.*";
```
emersion: Ditto

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+		}
+
+		for (size_t i = 0; i < found_count; ++i) {
+			char *relative_path = icon_glob.gl_pathv[i];
+
+			// Find the end of the current search path and walk to the next
+			// path component. Hopefully this will be the icon resolution
+			// subdirectory.
+			relative_path += strlen(theme_path);
+			while (*relative_path == '/') {
```
emersion: `relative_path[0]`

icon.c
```diff
@@ -28,7 +37,144 @@ static double fit_to_square(int width, int height, int square_size) {
...
+			break;
+		}
+		theme_path = strtok_r(NULL, ":", &saveptr);
+	}
+
+	if (icon_path == NULL) {
+		// Finally, fall back to looking in /usr/share/pixmaps. These are
+		// unsized icons, which may lead to downscaling, but some apps are
+		// still using it.
+		static const char *pixmaps_fmt = "/usr/share/pixmaps/%s.*";
```
emersion: Ditto

emersion:
This needs a rebase

vilhalmer:
Sorry I haven't finished fixing this up yet, I'll try to get to it later today so it's out of the way.

icon.c
vilhalmer: Good catch, I think I was originally planning to loop back and add an expansion of some sort and then forgot.

vilhalmer:
Alright! Review items addressed, rebased, and squished up a little nicer while I was at it. :)

emersion:
Thanks!

PR #129, opened 2019-03-25
Always print help when asked

vilhalmer:
Even if the config is broken. As a bonus, you can now also get up to two errors at a time: one from the config file, and one from an unrecognized cli flag.

There was also a memory leak here when printing help, though it didn't matter much since that immediately exits.

### DIFF
--- a/config.c
+++ b/config.c
@@ -662,16 +662,18 @@ int reload_config(struct mako_config *config, int argc, char **argv) {
 	struct mako_config new_config = {0};
 	init_default_config(&new_config);
 
-	if (load_config_file(&new_config) != 0) {
-		fprintf(stderr, "Failed to reload config\n");
+	int config_status = load_config_file(&new_config);
+	int args_status = parse_config_arguments(&new_config, argc, argv);
+
+	if (args_status > 0) {
+		finish_config(&new_config);
+		return args_status;
+	} else if (config_status < 0 || args_status < 0) {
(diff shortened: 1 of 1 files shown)


config.c
```diff
@@ -662,16 +662,19 @@ int reload_config(struct mako_config *config, int argc, char **argv) {
 	struct mako_config new_config = {0};
 	init_default_config(&new_config);
 
-	if (load_config_file(&new_config) != 0) {
-		fprintf(stderr, "Failed to reload config\n");
+	int config_status = load_config_file(&new_config);
+	int args_status = parse_config_arguments(&new_config, argc, argv);
+
+	if (args_status > 0) {
```
emersion: The check was `!=`. I'd prefer to keep it that way since it's a more idiomatic way to check for errors in C.

config.c
```diff
@@ -662,16 +662,19 @@ int reload_config(struct mako_config *config, int argc, char **argv) {
...
+		finish_config(&new_config);
+		return args_status;
+	}
+	else if (config_status < 0 || args_status < 0) {
```
emersion: Style: no newline between `}` and `else if`

config.c
emersion: Ah, I see. Okay, nevermind :P

config.c
vilhalmer: I've gotta add something to my editor to catch this.

emersion:
Thanks!

PR #131, opened 2019-03-31
Print usage message when not enough or too many argument are given on…

bobsaintcool:
Not a killer feature but could be nice to remove the following case:

```
$ /usr/bin/makoctl
makoctl: unrecognized command ''
```

The proposed patch offer to check how many argument are given from the CLI and directly print an error message with the standard usage function.


emersion:
Thanks!

PR #132, opened 2019-03-31
Add zsh completion function

bobsaintcool:
Did experiment with ZSH completion function. This one is not bullet proof but may help or get other started!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,6 +14,8 @@ project(
 add_project_arguments('-Wno-unused-parameter', language: 'c')
 add_project_arguments('-Wno-missing-braces', language: 'c')
 
+datadir = get_option('datadir')
+
 mako_inc = include_directories('include')
 
 cc = meson.get_compiler('c')
@@ -107,4 +109,18 @@ if scdoc.found()
 			install_dir: '@0@/man@1@'.format(mandir, section)
 		)
 	endforeach
+endif
+
+if get_option('zsh-completions')
+	zsh_files = files(
+		'completions/zsh/_makoctl',
+		'completions/zsh/_mako'
+	)
+	zsh_install_dir = datadir + '/zsh/site-functions'
+
+	install_data(
+		zsh_files,
+		install_dir: zsh_install_dir,
+		install_mode: 'rw-r--r--'
+	)
 endif
--- a/completions/zsh/_mako
+++ b/completions/zsh/_mako
@@ -1,0 +1,29 @@
+#compdef mako
+
+# Usage: mako [options...]
+
+
+# Colors can be specified with the format #RRGGBB or #RRGGBBAA.
+
+_arguments \
+    '-h[Show help message and quit.]' '--help[Show help message and quit.]' \
+    '--font[Font family and size.]:font:' \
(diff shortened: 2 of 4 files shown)


emersion:
Can you add it to Meson? (See how sway does it if you need inspiration)

bobsaintcool:
Will do, however allow me few day for that.

bobsaintcool:
Sorry for the delay I need a bit more time for getting into Meson. As its my first work with this don't hesitate if you would like some change.

I've also added a completion system for the mako command.

meson.build
```diff
@@ -107,3 +112,16 @@ if scdoc.found()
 		)
 	endforeach
 endif
+
+if get_option('zsh-completions')
+	zsh_files = files(
+		'completions/zsh/_makoctl',
+                'completions/zsh/_mako'
```
foxcpp: Mixed indent here.

bobsaintcool: I've should have fixed this issue in 05758b6298f6c69028fbb961b44d7a01ccdb8fd4 by using editorconfig, thx.

meson.build
```diff
@@ -107,3 +112,16 @@ if scdoc.found()
...
+
+if get_option('zsh-completions')
+	zsh_files = files(
+		'completions/zsh/_makoctl',
+		'completions/zsh/_mako'
+	)
+	zsh_install_dir = datadir + '/zsh/site-functions'
+
+	install_data(zsh_files,
+			install_dir: zsh_install_dir,
```
emersion: Extra indent here

bobsaintcool: Should 3b6603f4dd0528d5fbf7a8c5844c26edb4d4ff52 fix this

emersion:
Can you squash everything into a single commit?

emersion:
It seems like there are 3 commits now, including one unrelated commit.

You'll need to make sure you have this repo as a remote

```
git remote add upstream git@github.com:emersion/mako.git
```

Then rebase your PR

```
git fetch upstream
git rebase upstream/master
```

bobsaintcool:
I did an extra rebase and I think we are in a better state now, with only one commit for the above changes.

meson.build
```diff
@@ -14,6 +14,11 @@ project(
 add_project_arguments('-Wno-unused-parameter', language: 'c')
 add_project_arguments('-Wno-missing-braces', language: 'c')
 
+datadir = get_option('datadir')
+prefix = get_option('prefix')
+config = configuration_data()
```
emersion: `config` isn't used anywhere, and can be removed

emersion:
>I did an extra rebase and I think we are in a better state now, with only one commit for the above changes.

Indeed, looks much better.

Pro tip: to change the last commit, use `git commit --amend`.

meson.build
bobsaintcool: addressed in bf442f0159c87de04e18429b3e5da123680a7069

bobsaintcool:
> Pro tip: to change the last commit, use `git commit --amend`.

Will you change something in the squashed commit message?

emersion:
Thanks!

>Will you change something in the squashed commit message?

For small fixes, probably not

PR #133, opened 2019-04-01
Update README.md with dbus dependency and elogind tip

DarkDefender:
I've added two things to the readme that would have saved me a lot of time when I tried to get mako up and running on my system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ A lightweight notification daemon for Wayland. Works on Sway.
 If you're using Sway you can start mako on launch by putting `exec mako` in
 your configuration file.
 
+If you are using elogind, you might need to manually start a dbus user session:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +23,7 @@ Install dependencies:
 * pango
 * cairo
 * systemd or elogind (for the sd-bus library)
+* dbus (with user-session support)
```
emersion: Isn't this a transitive dependency of systemd/elogind?

DarkDefender: Not on Gentoo at least. I could install and run sway with elogind without any issues when dbus didn't have user-session support. (Systemd doesn't require user session support on dbus either)

It was only when I tried to run mako that I ran into issues. And it took quite a while to figure out what I needed to do, so having this would have been very useful to me.

emersion: That's true.

emersion:
Thanks!

Issue #102, opened 2019-01-03
Add border radius style option


0xdec:
It would be great to have a border-radius style option, so that we can have rounded rectangle notifications.

emersion:
See https://www.cairographics.org/samples/rounded_rectangle/

emersion:
Fixed in https://github.com/emersion/mako/pull/134

PR #134, opened 2019-04-01
Add border-radius option

esawady:
Added a new style option, border-radius, as per #102.

Progress bars are also rounded, and the notification is padded by at lest border-radius on the left and right.

### DIFF
--- a/config.c
+++ b/config.c
@@ -18,7 +18,6 @@ static int32_t max(int32_t a, int32_t b) {
 	return (a > b) ? a : b;
 }
 
-
 void init_default_config(struct mako_config *config) {
 	wl_list_init(&config->criteria);
 	struct mako_criteria *new_criteria = create_criteria(config);
@@ -93,6 +92,7 @@ void init_default_style(struct mako_style *style) {
 	style->padding.left = 5;
 
 	style->border_size = 2;
+	style->border_radius = 0;
 
 #ifdef HAVE_ICONS
 	style->icons = true;
@@ -254,6 +254,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	if (style->spec.invisible) {
 		target->invisible = style->invisible;
 		target->spec.invisible = true;
+	}
+
+	if (style->border_radius) {
+		target->border_radius = style->border_radius;
+		target->spec.border_radius = true;
 	}
 
 	return true;
@@ -429,7 +434,12 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	} else if (strcmp(name, "margin") == 0) {
 		return spec->margin = parse_directional(value, &style->margin);
 	} else if (strcmp(name, "padding") == 0) {
-		return spec->padding = parse_directional(value, &style->padding);
+		spec->padding = parse_directional(value, &style->padding);
+		if (spec->border_radius && spec->padding) {
+			style->padding.left = max(style->border_radius, style->padding.left);
+			style->padding.right = max(style->border_radius, style->padding.right);
+		}
+		return spec->padding;
 	} else if (strcmp(name, "border-size") == 0) {
 		return spec->border_size = parse_int(value, &style->border_size);
 	} else if (strcmp(name, "border-color") == 0) {
@@ -465,6 +475,13 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 			parse_criteria_spec(value, &style->group_criteria_spec);
 	} else if (strcmp(name, "invisible") == 0) {
 		return spec->invisible = parse_boolean(value, &style->invisible);
+	} else if (strcmp(name, "border-radius") == 0) {
+		spec->border_radius = parse_int(value, &style->border_radius);
+		if (spec->border_radius && spec->padding) {
+			style->padding.left = max(style->border_radius, style->padding.left);
+			style->padding.right = max(style->border_radius, style->padding.right);
+		}
+		return spec->border_radius;
 	}
 
 	return false;
--- a/include/config.h
+++ b/include/config.h
@@ -23,9 +23,9 @@ enum mako_sort_criteria {
 // fields in the mako_style structure should have a counterpart here. Inline
 // structs are also mirrored.
 struct mako_style_spec {
-	bool width, height, margin, padding, border_size, font, markup, format,
-		 actions, default_timeout, ignore_timeout, icons, max_icon_size,
-		 group_criteria_spec, invisible;
+	bool width, height, margin, padding, border_size, border_radius, font,
+		markup, format, actions, default_timeout, ignore_timeout, icons,
+		max_icon_size, group_criteria_spec, invisible;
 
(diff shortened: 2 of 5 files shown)


config.c
```diff
@@ -465,7 +480,14 @@ static bool apply_style_option(struct mako_style *style, const char *name,
...
 	} else if (strcmp(name, "invisible") == 0) {
 		return spec->invisible = parse_boolean(value, &style->invisible);
-	}
+	} else if (strcmp(name, "border-radius") == 0) {
+		spec->border_radius = parse_double(value, &style->border_radius);
+		if (spec->border_radius && spec->padding) {
+			style->padding.left = maxd(style->border_radius, style->padding.left);
+			style->padding.right = maxd(style->border_radius, style->padding.right);
+		}
+          	return spec->border_radius;
```
emersion: There are indentation issues here, we use tabs

emersion:
Could you squash all of these commits into one, and associate them with your GitHub account (by setting correctly the e-mail address)?

types.c
```diff
@@ -36,6 +36,13 @@ bool parse_int(const char *string, int *out) {
 	return errno == 0 && end[0] == '\0';
 }
 
+bool parse_double(const char *string, double *out) {
+	errno = 0;
+	char *end;
+	*out = (double)strtof(string, &end);
```
emersion: This should use `strtod`

render.c
```diff
@@ -197,6 +216,7 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 		draw_icon(cairo, icon, xpos, ypos, scale);
 	}
 
+	printf("h:%f,w:%f\n", offset_x + style->border_size + text_x, offset_y + style->border_size + (double)(notif_height - border_size - text_height) / 2);
```
emersion: Probably a remnant from your debugging sessions :)

render.c
```diff
@@ -36,9 +37,32 @@ static void move_to(cairo_t *cairo, double x, double y, int scale) {
 	cairo_move_to(cairo, x * scale, y * scale);
 }
 
-static void set_rectangle(cairo_t *cairo, double x, double y, double width, double height,
-		int scale) {
-	cairo_rectangle(cairo, x * scale, y * scale, width * scale, height * scale);
+static void rounded_rectangle(cairo_t *cairo, double x, double y, double width,
+                    double height, double radius) {
```
emersion: Style: indent this line with two tabs

render.c
```diff
@@ -13,6 +13,7 @@
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 #include "icon.h"
 
+#define M_PI 3.1415926535897932384
```
emersion: Pretty sure we can get this from `<math.h>`. IIRC the magic incantation is to replace `#define _POSIX_C_SOURCE 200809L` with `#define _XOPEN_SOURCE 700`.

include/config.h
```diff
@@ -23,9 +23,9 @@ enum mako_sort_criteria {
 // fields in the mako_style structure should have a counterpart here. Inline
 // structs are also mirrored.
 struct mako_style_spec {
-	bool width, height, margin, padding, border_size, font, markup, format,
-		 actions, default_timeout, ignore_timeout, icons, max_icon_size,
-		 group_criteria_spec, invisible;
+	bool width, height, margin, padding, border_size, border_radius, font,
+		 markup, format, actions, default_timeout, ignore_timeout, icons,
```
emersion: Extra space after the indentation

render.c
```diff
@@ -36,9 +37,32 @@ static void move_to(cairo_t *cairo, double x, double y, int scale) {
...
+	cairo_arc(cairo, x + width - radius, y + height - radius, radius, 0 * degrees, 90 * degrees);
+	cairo_arc(cairo, x + radius, y + height - radius, radius, 90 * degrees, 180 * degrees);
+	cairo_arc(cairo, x + radius, y + radius, radius, 180 * degrees, 270 * degrees);
+	cairo_close_path(cairo);
+} 
+
+static void set_rounded_rectangle(cairo_t *cairo, double x, double y, double width, double height,
+		int scale, double radius) {
+	if (width != 0 && height != 0) {
+		rounded_rectangle(cairo, x * scale, y * scale, width * scale,
```
emersion: This could just be

```c
if (width == 0 || height == 0) {
	return;
}
```

at the beginning of the function, no need to declare two functions

esawady:
Thanks!

One other question: I've been noticing some memory leaks (`valgrind build/mako`), however there are memory leaks in master as well. Should I be worrying about that or is that a separate issue?

main.c
```diff
@@ -28,6 +28,7 @@ static const char usage[] =
 	"                                  meaning as in CSS.\n"
 	"      --border-size <px>          Border size.\n"
 	"      --border-color <color>      Border color.\n"
+	"      --border-radius <px>        Corner radius\n"
```
vilhalmer: If this is described as pixels, does it really make sense to parse it as a double? It's also used as an int later on which leads me to believe that it doesn't need to be a float.

render.c
```diff
@@ -36,9 +36,31 @@ static void move_to(cairo_t *cairo, double x, double y, int scale) {
...
+		return;
+	}
+	x *= scale;
+	y *= scale;
+	width *= scale;
+	height *= scale;
+	double degrees = M_PI / 180.0;
+
+	if (radius > width / 2) {
+		radius = 0;
```
vilhalmer: Should these snap to `width / 2` (or height) rather than throwing it out entirely?

render.c
```diff
@@ -78,6 +100,7 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	int border_size = 2 * style->border_size;
 	int padding_height = style->padding.top + style->padding.bottom;
 	int padding_width = style->padding.left + style->padding.right;
+        int radius = style->border_radius;
```
vilhalmer: This type doesn't match.

render.c
esawady: I prefer to widen the rectangle to accomodate the radius (I had done this in the past, it must have been undone at some point). This should only come up in very thin progress bars, and I think it's better to have the progress bar be slightly wider than expected than to render it with the wrong radius (which would leave a sliver of the background color in between the border and the progress bar).

config.c
```diff
@@ -93,6 +92,7 @@ void init_default_style(struct mako_style *style) {
 	style->padding.left = 5;
 
 	style->border_size = 2;
+        style->border_radius = 0;
```
emersion: Style: we use tabs

config.c
```diff
@@ -256,6 +256,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 		target->spec.invisible = true;
 	}
 
+        if (style->border_radius) {
```
emersion: Style: we use tabs (applies to other lines you've modified too)

emersion:
Thanks!

PR #135, opened 2019-04-03
AppArmor profile

CameronNemo:
Signed-off-by: Cameron Nemo <camerontnorman@gmail.com>


foxcpp:
Out of curiosity, why does it require `/{run,dev}/shm/grim-*`?

CameronNemo:
I couldn't tell you, beyond 'wayland'...

vilhalmer:
`grim` being used as the shm filename was a copy-paste incident which has been resolved on master as far as I know. It should use `mako-*` now. The files themselves are necessary because we don't have a POSIX-y way of opening anonymous shared memory to push frames to the compositor.

apparmor/fr.emersion.Mako
```diff
@@ -0,0 +1,22 @@
...
+  #include <abstractions/dbus-strict>
+  dbus bind
+       bus=session
+       name=org.freedesktop.Notifications,
+  dbus receive
+       bus=session
+       path=/fr/emersion/Mako
+       interface=fr.emersion.Mako,
+
+  /{run,dev}/shm/grim-* rw,
```
vilhalmer: This should be `mako-*` now on master.

CameronNemo: Thanks for the information... I am using the release. Fixed now.

CameronNemo:
>The files themselves are necessary because we don't have a POSIX-y way of opening anonymous shared memory to push frames to the compositor

There could be an abstraction for memfd and shm files available. Not sure if it is MIT.

emersion:
We use `shm_open`, so the way we create in-memory files for our buffers is POSIX.

apparmor/fr.emersion.Mako
```diff
@@ -0,0 +1,22 @@
...
+  #include <abstractions/dbus-strict>
+  dbus bind
+       bus=session
+       name=org.freedesktop.Notifications,
+  dbus receive
+       bus=session
+       path=/fr/emersion/Mako
+       interface=fr.emersion.Mako,
+
+  /{run,dev}/shm/mako-* rw,
```
emersion: `/run/shm` doesn't exist for me.

emersion:
Are there other standards like AppArmor?

apparmor/fr.emersion.Mako
CameronNemo: Right, it is fairly niche. Could still pop up.

emersion: Where is it coming from?

CameronNemo:
@emersion not sure what you mean by standards. There is SELinux, which is much more broadly deployed (Android, Chrome OS, CentOS/RHEL). There is also SMACK, Tomoyo, and grsec's RBAC. Those are far more niche than AppArmor. AA has been enabled by default on Ubuntu and openSUSE/SLES for a long time (~ ten years). Debian will enable it by default for the upcoming Buster release. It is available as an opt in on many other distros.

emersion:
Okay, thanks. I was just wondering if there were other widely used alternatives.

apparmor/fr.emersion.Mako
CameronNemo: https://github.com/zfsonlinux/mountall/blob/57619b318090a4226c69d24d035e6568ec2ca0c9/src/fstab#L23

CameronNemo: https://wiki.debian.org/ReleaseGoals/RunDirectory#Packages_using_.2Fdev.2Fshm

emersion: Oh okay, so it's the thing that will replace `/dev/shm`. Let's keep it.

emersion:
What about Polkit?

CameronNemo:
As an alternative to AppArmor, or does mako use polkit currently and the AA profile does not account for that? If as an alternative, they do very different jobs. AppArmor only ever reduces the permissions granted to a process, while polkit allows for targeted privilege escalation.

emersion:
Eh, right, I wonder why I was confused.

What held me off is: should we (the authors) provide a policy file? I'd expect this to be the package maintainers' job, so that we don't add too many privileges in the policy.

Another concern is that it adds files specific to one policy mechanism. Should we merge systemd services too for instance?

CameronNemo:
I would actually be concerned that there would be too few priveleges in the policy. Imagine for example that a new feature is added. An unaware package maintainer may fail to update the LSM policy to account for that. Furthermore splitting the responsibility up to several non-communicating package maintainers could create more work.

As for a systemd user service, I think it is totally fair game. As long as it is a build option like this it should create no issues. Especially since users would need to explicitly enable the service, so it would be unlikely to conflict with other methods of starting and supervising mako.

emersion:
Well the downside for us is that we'll need to maintain the file. If we get a unit service for each init system plus a bunch of files for each distribution's policy mechanism, this would start to get annoying (I don't know each file format/semantics, I can't test easily if it works, and so on).

OTOH we also have a PR for [shell completions](https://github.com/emersion/mako/pull/132).

All in all, there's no perfect solution for everybody. Maybe we could put all these files in e.g. a `contrib/` directory to make it clear we won't _necessarily_ keep them up-to-date ("please fix them if they are broken")?

(Note: it seems easy enough to update the AppArmor policy, but the issue is remember we have an AppArmor policy and realizing we need to update it -- as I'm not an AppArmor user, it won't break anything for me)

foxcpp:
>(Note: it seems easy enough to update the AppArmor policy, but the issue is remember we have an AppArmor policy and realizing we need to update it -- as I'm not an AppArmor user, it won't break anything for me)

I am [AppArmor user](https://github.com/foxcpp/apparmor-profiles)! Can I be useful here?

emersion:
>I am AppArmor user! Can I be useful here?

Well yes! Report issues if you find something's not working :)

emersion:
Thanks!

PR #137, opened 2019-04-10
Add a missing command line border-radius option

Difrex:
Hi,

I just compiled the latest master source and `--border-radius` option does not work.
I launch mako with these command line options below and got an error:
```
mako --border-radius 5
mako: unrecognized option '--border-radius'
Failed to parse config
```

In this PR I just added a missing command line `border-radius` option.


emersion:
Thanks!

PR #139, opened 2019-04-13
icon.c: fix build without icons support

xdavidwu:
Fix conflicting types for `create_icon` if build with icons support disabled.


emersion:
Thanks!

(I was about to add a build without icons to the CI, but I'm hitting https://gitlab.gnome.org/GNOME/pango/issues/344)

PR #143, opened 2019-04-17
icon: add support for icon size and scale in second level subdir name

xdavidwu:
Some icon themes (like Arc, breeze ...) have their icon size and scale attribute in second level subdir name (paths like actions/16/add-files-to-archive.png). This pull request add support to this kind of naming.

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -112,7 +112,16 @@ static char *resolve_icon(struct mako_notification *notif) {
 			errno = 0;
 			int32_t icon_size = strtol(relative_path, NULL, 10);
 			if (errno || icon_size == 0) {
-				continue;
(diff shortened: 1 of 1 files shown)


icon.c
```diff
@@ -112,7 +112,15 @@ static char *resolve_icon(struct mako_notification *notif) {
 			errno = 0;
 			int32_t icon_size = strtol(relative_path, NULL, 10);
 			if (errno || icon_size == 0) {
-				continue;
+				errno = 0;
```
vilhalmer: Can you add a comment here explaining what this is looking for?

xdavidwu: Done.

vilhalmer:
Thanks!

PR #144, opened 2019-04-17
parse_config_arguments: add missing icon-path option

xdavidwu:
icon-path is missing from getopt_long options.


emersion:
Thanks!

PR #147, opened 2019-04-26
Add image-data hint support

xdavidwu:
Add image-data hint support, which is preferred when both image-data and app_icon exist.

Tested on Firefox (which uses libnotify).

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -217,6 +217,79 @@ static int handle_notify(sd_bus_message *msg, void *data,
 				return ret;
 			}
 			notif->progress = progress;
+		} else if (strcmp(hint, "image-data") == 0) {
+			ret = sd_bus_message_enter_container(msg, 'v', "(iiibiiay)");
+			if (ret < 0) {
+				return ret;
+			}
+
+			ret = sd_bus_message_enter_container(msg, 'r', "iiibiiay");
+			if (ret < 0) {
+				return ret;
+			}
+
+			struct mako_image_data *image_data = calloc(1, sizeof(struct mako_image_data));
+			if (image_data == NULL) {
+				return -1;
+			}
+
+			ret = sd_bus_message_read(msg, "iiibii", &image_data->width,
+					&image_data->height, &image_data->rowstride,
+					&image_data->has_alpha, &image_data->bits_per_sample,
+					&image_data->channels);
+			if (ret < 0) {
+				free(image_data);
+				return ret;
+			}
+
+			// Calculate the expected useful data length without padding in last row
+			// len = size before last row + size of last row
+			//     = (height - 1) * rowstride + width * ceil(channels * bits_pre_sample / 8.0)
+			size_t image_len = (image_data->height - 1) * image_data->rowstride +
+				image_data->width * ((image_data->channels *
+				image_data->bits_per_sample + 7) / 8);
+			uint8_t *data = calloc(image_len, sizeof(uint8_t));
+			if (data == NULL) {
+				free(image_data);
+				return -1;
+			}
+
+			ret = sd_bus_message_enter_container(msg, 'a', "y");
+			if (ret < 0) {
+				free(data);
+				free(image_data);
+				return ret;
+			}
+
+			// Ignore the extra padding bytes in the last row if exist
+			for (size_t index = 0; index < image_len; index++) {
+				uint8_t tmp;
+				ret = sd_bus_message_read(msg, "y", &tmp);
+				if (ret < 0){
+					free(data);
+					free(image_data);
+					return ret;
+				}
+				data[index] = tmp;
+			}
+
+			image_data->data = data;
+			notif->image_data = image_data;
+
+			ret = sd_bus_message_exit_container(msg);
+			if (ret < 0) {
+				return ret;
+			}
+
+			ret = sd_bus_message_exit_container(msg);
+			if (ret < 0) {
+				return ret;
+			}
+
+			ret = sd_bus_message_exit_container(msg);
+			if (ret < 0) {
+				return ret;
+			}
 		} else {
 			ret = sd_bus_message_skip(msg, "v");
 			if (ret < 0) {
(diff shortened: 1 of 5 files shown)


include/icon.h
```diff
@@ -11,6 +11,16 @@ struct mako_icon {
 	cairo_surface_t *image;
 };
 
+struct iiibiiay {
```
emersion: Eh. `aiii(siosssss(si)iiiiii)iiiba{yys}a`?

(Translation from D-Bus to English: can we change this to a human-readable name?)

dbus/xdg.c
```diff
@@ -217,6 +217,79 @@ static int handle_notify(sd_bus_message *msg, void *data,
...
+					&image_data->has_alpha, &image_data->bits_per_sample,
+					&image_data->channels);
+			if (ret < 0) {
+				free(image_data);
+				return ret;
+			}
+
+			size_t image_len = (image_data->height - 1) * image_data->rowstride +
+				image_data->width * ((image_data->channels *
+				image_data->bits_per_sample + 7) / 8);
```
emersion: Hmm. I'm not sure I get this formula.

dbus/xdg.c
```diff
@@ -217,6 +217,79 @@ static int handle_notify(sd_bus_message *msg, void *data,
...
+				uint8_t tmp;
+				ret = sd_bus_message_read(msg, "y", &tmp);
+				if (ret == 0) {
+					break;
+				} else if (ret < 0 || index >= image_len) {
+					free(data);
+					free(image_data);
+					return ret;
+				}
+				data[index] = tmp;
```
emersion: Can we add a length check just before this?

icon.c
```diff
@@ -184,16 +195,24 @@ static char *resolve_icon(struct mako_notification *notif) {
...
-	if (path == NULL) {
-		return NULL;
+	GdkPixbuf *image = NULL;
+	if (notif->image_data != NULL) {
+		image = load_image_data(notif->image_data);
 	}
 
-	GdkPixbuf *image = load_image(path);
-	free(path);
 	if (image == NULL) {
```
emersion: `} else {`

xdavidwu: I think we should fallback to app_icon if loading image-data fails?

include/icon.h
xdavidwu: Not sure what I should name it. Any suggestion?

dbus/xdg.c
xdavidwu: I've added some comments in code to explain that.

dbus/xdg.c
xdavidwu: Not sure what you mean. Out-of-range check is already there. `} else if (ret < 0 || index >= image_len) {`

include/icon.h
emersion: `mako_icon_data`?

dbus/xdg.c
emersion: Oh, right, missed that

icon.c
emersion: Okay

include/icon.h
xdavidwu: Done!

dbus/xdg.c
emersion: Why is the last row special?

xdavidwu: > If you are doing memcpy() of raw pixbuf data, note that the last row in the pixbuf may not be as wide as the full rowstride, but rather just as wide as the pixel data needs to be. That is, it is unsafe to do memcpy (dest, pixels, rowstride * height) to copy a whole pixbuf. Use gdk_pixbuf_copy() instead, or compute the width in bytes of the last row as width * ((n_channels * bits_per_sample + 7) / 8).
(Quote from GdkPixbuf)

GdkPixBuf may trim the last row to save space.
Libnotify and notification-daemon also use this formula.
Not sure if there is any client implementation that send untrimmed data, maybe we should just ignore extra bits instead of an error?

emersion: >GdkPixBuf may trim the last row to save space.

I don't understand how they can do so. If the image contains 1000 pixels, you can't "trim the last row" and only send 998 pixels.

>maybe we should just ignore extra bits instead of an error?

Or do you mean they pad the image data (adding extra bytes we shouldn't care about)?

In this case we should just ignore extra bytes, yes.

xdavidwu: Yes I meant extra padding bytes may exist in the last row.

xdavidwu: I've updated the commit to ignore extra bytes, also changed the comment style to fit with other comments.

emersion: Then this formula can be simplified to `image_data->height * image_data->rowstride`?

dbus/xdg.c
```diff
@@ -217,6 +217,85 @@ static int handle_notify(sd_bus_message *msg, void *data,
...
+				uint8_t tmp;
+				ret = sd_bus_message_read(msg, "y", &tmp);
+				if (ret == 0) {
+					break;
+				} else if (ret < 0){
+					free(data);
+					free(image_data);
+					return ret;
+				}
+				// Ignore the extra padding bytes in the last row if exist
```
emersion: Put do we need to `sd_bus_message_read` all elements? Can't we just `sd_bus_message_exit_container` to discard them?

xdavidwu: Updated.

dbus/xdg.c
xdavidwu: `image_data->rowstride` contains the extra padding bytes, so`image_data->height * image_data->rowstride` contains the extra bytes in the last row. Client may not send extra bytes in last row, so we should expect the length to be without that.

emersion: In other words: the last row may not be padded to `rowstride`?

Gah, this sucks. Okay, let's just keep this version.

emersion:
Thanks for your contribution!

PR #148, opened 2019-05-04
Add icon_data hint support

joelanford:
Building on #147, this adds support for the deprecated (though still in use) `icon_data` hint, by simply adding another `strcmp()` call to the condition that checks for `image-data` so that either will be used.

My motivation for adding this is that the Linux Spotify client is still using `icon_data` hints in its notifications.


emersion:
Thanks!

PR #149, opened 2019-05-04
Fix memory leak if there are multiple icon data

xdavidwu:
Without this patch, memory leak can happen if both image-data and icon_data are present, or if there are hint entries with duplicate key that is either image-data or icon_data.


emersion:
Thanks!

Issue #121, opened 2019-03-07
Add a makoctl command to list current notifications


StephenBrown2:
I wanted to set up some criteria, as I had with dunst, and I would like to see what the criteria of the notifications are so I can match them more easily.

dunst has a `-print` flag that will output useful information about a notification to write proper rules. As far as I can tell, mako has no such option (yet).

Src[1]: https://wiki.archlinux.org/index.php/Dunst#Filtering
Src[2]: https://github.com/dunst-project/dunst/blob/master/dunstrc#L337-L338

emersion:
Maybe we should instead add a `makoctl` command to get a list of current notifications with all metadata. Could be more useful than just writing criteria easily.

PR #150, opened 2019-05-04
Add ListNotifications to D-Bus API

emersion:
TODO:
- [x] Update docs

Fixes https://github.com/emersion/mako/issues/121

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -69,6 +69,72 @@ done:
 	return sd_bus_reply_method_return(msg, "");
 }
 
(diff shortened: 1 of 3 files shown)


vilhalmer:
lgtm! I could see including some more of the internal state as well, but we can wait and see what ends up being necessary/useful.

emersion:
Yup. I intentionally only included a few fields because it can be easily extended as needed.

Updated with docs.

vilhalmer:
Thanks!

Issue #152, opened 2019-05-08
Criteria hierarchy and examples


leachim:
I would like to give different notifications styles to criteria, based on where the notifications originates from. Concrete example: I am using multiple slack work spaces, each of them lives in a different browser tab. Now, when I receive a notification in a specific work space I would like to receive a permanent, sticky notification so I don't miss it. But not for all of the work spaces. 

How can I filter messages from a specific work space? In general, is there a good way to find out the fields/criteria available from a notification? 
Would love to see some more example configurations to see how people handle this in practice for common desktop apps.

StephenBrown2:
I've enabled exact string matching with this diff, as I have similar requirements (want notifications from a particular user to stand out):
```diff
diff --git a/criteria.c b/criteria.c
index 907b09e..ae6dc8c 100644
--- a/criteria.c
+++ b/criteria.c
@@ -262,8 +262,12 @@ bool apply_criteria_field(struct mako_criteria *criteria, char *token) {
                        }
                        criteria->spec.group_index = true;
                        return true;
+               } else if (strcmp(key, "summary") == 0) {
+                       criteria->summary = strdup(value);
+                       criteria->spec.summary = true;
+                       return true;
                } else {
-                       // TODO: summary + body, once we support regex and they're useful.
+                       // TODO: body, once we support regex and they're useful.
                        // Anything left must be one of the boolean fields, defined using
                        // standard syntax. Continue on.
                }
```

My config for that looks like:
```
[summary="New message from Important User"]
background-color=#C40023
border-color=#EB000
```

I don't think we can set _urgency_ or disable _timeout_ via criteria yet, but the background change is a good start for me at least.

vilhalmer:
@StephenBrown2 If you post that as a PR I'd be +1 on it, it's really simple. :) Alternately, I don't think adding regex is actually that difficult. We should use whatever library sway is using for their criteria.

Urgency will never be settable, allowing criteria sections to modify fields provided by the D-Bus spec creates an need for effectively unbounded re-application of criteria to converge on a final style. In better news, you can ignore timeouts via `ignore-timeout=true`.

CameronNemo:
Does sway use regex or globbing? The latter just needs [fnmatch()](https://pubs.opengroup.org/onlinepubs/009695399/functions/fnmatch.html).

ascent12:
IIRC sway uses PCRE for its criteria (because that's what i3 uses). You could get away with simply using POSIX Extended Regular Expressions here, though.

[...]

PR #153, opened 2019-05-13
Add exact-string matching to criteria

StephenBrown2:
Regarding my comment on https://github.com/emersion/mako/issues/152#issuecomment-490964953, this implements exact string matching on the summary of a notification, for use with known notification titles, i.e. from messaging apps.


emersion:
Thanks!

PR #156, opened 2019-05-23
Add a D-Bus service file for mako

meiyopeng:
This patch adds a D-Bus service file for mako.

I have never written any meson.build file before.  Suggestions are welcome.

This patch also solves an annoying issue for [Guix System](https://www.gnu.org/software/guix/).  Guix has no systemd init system and $DBUS_SESSION_BUS_ADDRESS is not set in user sessions.  
we had to start mako with `dbus-launch --autolaunch=$(dbus-uuidgen --get) mako`.  After applying this patch, mako can be automatically started by the D-Bus daemon. That complex command is not needed anymore.  I will update the package in Guix when this patch is merged.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -92,6 +92,16 @@ install_data(
 	install_mode: 'rwxr-xr-x',
 )
 
+conf_data = configuration_data()
+conf_data.set('bindir', join_paths(get_option('prefix'), get_option('bindir')))
+
+configure_file(
+	configuration: conf_data,
+	input: 'fr.emersion.mako.service.in',
+	output: '@BASENAME@',
+	install_dir: datadir + '/dbus-1/services',
+)
+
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -92,6 +92,16 @@ install_data(
 	install_mode: 'rwxr-xr-x',
 )
 
+conf_data = configuration_data()
+conf_data.set('prefix', get_option('prefix'))
```
emersion: Can't we use `bindir` here?

meiyopeng: How about:
```
conf_data.set('bindir', join_paths(get_option('prefix'), get_option('bindir')))
```
And in fr.emersion.mako.service.in:
```
[D-BUS Service]
Name=org.freedesktop.Notifications
Exec=@bindir@/mako
```

emersion:
Thanks!

PR #158, opened 2019-06-01
Allow listing and invoking actions with makoctl

vilhalmer:
This is like 2% of the way toward #4. I'm posting this mostly to see if this method of interacting with actions is even useful, or if it's too difficult. With this change, you can invoke actions like:
```
makoctl invoke default 33
```
where `default` is the action's key and `33` is the internal id. I chose to expose the id rather than just using the visible index to avoid race conditions in external scripts. You still have to invoke the action before the notification vanishes, you can't accidentally invoke it on the wrong one this way.

Here's some example `jq` for listing the json of visible notifications, which contains the id and action list necessary to invoke this:
```
makoctl list | jq -r '.data[][] | .[] |= .data'
```

I may have done the sd-bus stuff in the hardest possible way, I'm not super familar with the library and there are few manpages.

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -46,19 +46,14 @@ static int handle_invoke_action(sd_bus_message *msg, void *data,
 	struct mako_state *state = data;
 
 	uint32_t id = 0;
-	int ret = sd_bus_message_read(msg, "u", &id);
+	const char *action_key;
+	int ret = sd_bus_message_read(msg, "us", &id, &action_key);
 	if (ret < 0) {
 		return ret;
 	}
 
 	if (id == 0) {
 		id = state->last_id;
-	}
-
-	const char *action_key;
-	ret = sd_bus_message_read(msg, "s", &action_key);
-	if (ret < 0) {
-		return ret;
 	}
 
 	if (wl_list_empty(&state->notifications)) {
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -140,24 +140,24 @@ static int handle_list_notifications(sd_bus_message *msg, void *data,
 			return ret;
 		}
 
-		// Actions are an array of strings.
-		ret = sd_bus_message_open_container(reply, 'v', "as");
+		ret = sd_bus_message_open_container(reply, 'v', "a{ss}");
 		if (ret < 0) {
 			return ret;
 		}
 
-		ret = sd_bus_message_open_container(reply, 'a', "s");
+		ret = sd_bus_message_open_container(reply, 'a', "{ss}");
 		if (ret < 0) {
 			return ret;
 		}
 
 		struct mako_action *action;
 		wl_list_for_each(action, &notif->actions, link) {
-			ret = sd_bus_message_append(reply, "s", action->key);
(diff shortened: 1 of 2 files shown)


dbus/mako.c
```diff
@@ -45,8 +45,18 @@ static int handle_invoke_action(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct mako_state *state = data;
 
+	uint32_t id = 0;
+	int ret = sd_bus_message_read(msg, "u", &id);
```
emersion: You can merge both `sd_bus_message_read` calls, with something among the lines of:

```
int ret = sd_bus_message_read(msg, "us", &id, &action_key);
```

makoctl
```diff
@@ -4,19 +4,21 @@ usage() {
 	echo "Usage: makoctl <command> [options...]"
 	echo ""
 	echo "Commands:"
-	echo "  dismiss [-a|--all] Dismiss the last or all notifications"
-	echo "  invoke [action]    Invoke an action on the last notification"
-	echo "  list               List notifications"
-	echo "  reload             Reload the configuration file"
-	echo "  help               Show this help"
+	echo "  dismiss [-a|--all]    Dismiss the last or all notifications"
+	echo "  invoke [action] [id]  Invoke an action on the notification"
```
emersion: Maybe we should use an option for this, having multiple optional arguments is a little bit hairy.

```
invoke [-n id] [action]
```

Thoughts?

dbus/mako.c
```diff
@@ -115,6 +129,55 @@ static int handle_list_notifications(sd_bus_message *msg, void *data,
...
+		}
+
+		ret = sd_bus_message_open_container(reply, 'a', "s");
+		if (ret < 0) {
+			return ret;
+		}
+
+		struct mako_action *action;
+		wl_list_for_each(action, &notif->actions, link) {
+			ret = sd_bus_message_append(reply, "s", action->key);
```
emersion: Should we keep this an array of keys, or should we also add the title too?

vilhalmer:
Wow, I totally forgot about this PR. Will get it finished up asap.

makoctl
vilhalmer: I like this idea.

dbus/mako.c
vilhalmer: I changed it to a dict of identifier -> title. I think I finally fully understand dbus types now. Incidentally, the fact that our json output appears to have an extra level of array nesting (like we had a confusing discussion about in IRC a while back) seems to be a busctl display bug, gdbus does not show the same outer structure.

emersion: Makes sense.

Clients can also add additional metadata to actions, like action-icons. Should we make actions a dict of identifier -> dict of action properties?

emersion: w/e, let's just break our API if we need to.

vilhalmer: Ah sorry I was busy yesterday. We'll just have to break it before we release again. :)

emersion: Oh, no worries! Just wanted to get this merged, no big deal to merge as-is :)

PR #160, opened 2019-06-02
Error for when users run Mako and it's already running.

qvm1Wak:
This tripped me up for a second, so maybe the message will help the next person.

### DIFF
--- a/dbus/dbus.c
+++ b/dbus/dbus.c
@@ -31,7 +31,11 @@ bool init_dbus(struct mako_state *state) {
 
 	ret = sd_bus_request_name(state->bus, service_name, 0);
 	if (ret < 0) {
-		fprintf(stderr, "Failed to acquire service name: %s. Is Mako already running?\n", strerror(-ret));
+                if (ret == -EEXIST) {
+                        fprintf(stderr, "Failed to acquire service name: %s. Is Mako already running?\n", strerror(-ret));
+                } else {
+                        fprintf(stderr, "Failed to acquire service name: %s\n", strerror(-ret));
+                }
 		goto error;
 	}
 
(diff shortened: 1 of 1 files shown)


dbus/dbus.c
```diff
@@ -31,7 +31,7 @@ bool init_dbus(struct mako_state *state) {
 
 	ret = sd_bus_request_name(state->bus, service_name, 0);
 	if (ret < 0) {
-		fprintf(stderr, "Failed to acquire service name: %s\n", strerror(-ret));
+		fprintf(stderr, "Failed to acquire service name: %s. Is Mako already running?\n", strerror(-ret));
```
emersion: Maybe we can check the value of `ret` and only print the new message in case it's already running?

vilhalmer:
I like this, but it should probably be "is another notification daemon running?" because it will show up if anything owns the name `org.freedesktop.Notifications`, not just mako.

dbus/dbus.c
```diff
@@ -31,7 +31,11 @@ bool init_dbus(struct mako_state *state) {
 
 	ret = sd_bus_request_name(state->bus, service_name, 0);
 	if (ret < 0) {
-		fprintf(stderr, "Failed to acquire service name: %s. Is Mako already running?\n", strerror(-ret));
+                if (ret == -EEXIST) {
```
emersion: Style: we use tabs

dbus/dbus.c
```diff
@@ -31,7 +31,11 @@ bool init_dbus(struct mako_state *state) {
...
+                        fprintf(stderr, "Failed to acquire service name: %s. Is Mako already running?\n", strerror(-ret));
```
emersion: We can probably always print the existing log line, but print an additional one after if `ret == -EEXIST`, saying the another notification daemon is already running.

emersion:
Anyway, this looks good to me as-is

emersion:
Thanks!

PR #166, opened 2019-06-27
Properly URL decode file:// URIs

sauyon:
I wasn't sure where to put the helper functions so I just dumped them before the function that used them.

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <assert.h>
+#include <ctype.h>
 #include <cairo/cairo.h>
 
 #include "mako.h"
@@ -49,6 +50,30 @@ static double fit_to_square(int width, int height, int square_size) {
 	return longest > square_size ? square_size/longest : 1.0;
 }
 
+static char hex_val(char digit) {
+	assert(isxdigit(digit));
+	if (digit >= 'a') {
+		return digit - 'a' + 10;
+	} else if (digit >= 'A') {
+		return digit - 'A' + 10;
+	} else {
+		return digit - '0';
+	}
+}
+
+static void url_decode(char *dst, const char *src) {
+	while (src[0]) {
+		if (src[0] == '%' && isxdigit(src[1]) && isxdigit(src[2])) {
+			dst[0] = 16*hex_val(src[1]) + hex_val(src[2]);
+			dst++; src += 3;
+		} else {
+			dst[0] = src[0];
+			dst++; src++;
+		}
+	}
+	dst[0] = '\0';
+}
+
 // Attempt to find a full path for a notification's icon_name, which may be:
 // - An absolute path, which will simply be returned (as a new string)
 // - A file:// URI, which will be converted to an absolute path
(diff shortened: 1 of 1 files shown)


icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
 	return longest > square_size ? square_size/longest : 1.0;
 }
 
+static char hexVal(char digit) {
```
emersion: Style: use `snake_case`

icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	assert(isxdigit(digit));
+	if (digit >= 'a') return digit - 'a' + 10;
```
emersion: Style: always use brackets, even for one-line `if` body

icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	else if (digit >= 'A') return digit - 'A' + 10;
+	else return digit - '0';
+}
+
+static void urldecode(char *dst, const char *src) {
+	while (*src) {
```
emersion: Prefer `src[0]` over `*src`

icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
...
+		if ((*src == '%') &&
+				(isxdigit(src[1]) && isxdigit(src[2]))) {
```
emersion: Style: unnecessary parentheses

icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
...
+			*dst++ = 16*hexVal(src[1]) + hexVal(src[2]);
+			src+=3;
```
emersion: Style: missing spaces around operator

icon.c
```diff
@@ -49,6 +50,26 @@ static double fit_to_square(int width, int height, int square_size) {
...
+	if (digit >= 'a') return digit - 'a' + 10;
+	else if (digit >= 'A') return digit - 'A' + 10;
+	else return digit - '0';
+}
+
+static void urldecode(char *dst, const char *src) {
+	while (*src) {
+		if ((*src == '%') &&
+				(isxdigit(src[1]) && isxdigit(src[2]))) {
+			*dst++ = 16*hexVal(src[1]) + hexVal(src[2]);
```
emersion: Please advance `dst` in a separate instruction

icon.c
```diff
@@ -67,8 +88,11 @@ static char *resolve_icon(struct mako_notification *notif) {
 		return strdup(icon_name);
 	}
 	if (strstr(icon_name, "file://") == icon_name) {
-		// Just chop off the scheme.
-		return strdup(icon_name + strlen("file://"));
+		// Chop off the scheme and URL decode
+		char *res = malloc(strlen(icon_name)+1-strlen("file://"));
```
emersion: Style: missing spaces around `+` and `-`

icon.c
```diff
@@ -67,8 +88,11 @@ static char *resolve_icon(struct mako_notification *notif) {
...
+		if (!res) return res;
```
emersion: Style: brackets for one-line condition body

sauyon:
Should have looked more carefully at the surrounding code; apologies. Was a quick fix I wrote up at work, and some of the (imo) ugliness from the [StackOverflow solution](https://stackoverflow.com/a/14530993) the code is based on was left over.

Re the fact that it's StackOverflow code I've modified it enough that I'm reasonably sure it doesn't need attribution at this point (and indeed there is no attribution required by default) but if you'd like I can add a comment or something.

sauyon:
Ok, finished my own style audit.

Is there a reason `assert.h` is included twice btw?

Emantor:
FWIW, code contributed before 01.02.2016 is CC-BY-SA which requires attribution. IANAL of course. Also open source licensing is sometimes a mess.

sauyon:
Ah, missed that. I’m still reasonably sure that especially at this point the code is different enough (and/or basic enough) that it should be fine, but obviously IANAL as well.

emersion:
>Is there a reason assert.h is included twice btw?

Ah, this is a mistake

Issue #167, opened 2019-06-28
Replace non-existent id


TheEdward162:
Hi,

recently I have started using `mako` instead of `dunst` and I noticed a difference in behaviour that I believe to be a bug.

If I call `notify-send.sh --print-id --replace 100 App Body`, the returned id is not 100 unless the notification with id 100 already exists. However, according to [the spec](http://www.galago-project.org/specs/notification/0.9/x408.html):

> If replaces_id is not 0, the returned value is the same value as replaces_id. 

the returned id should be the same as `--replace` id (no mention on whether or not the notification exists). Dunst does return the same id even for non-existent notifications.

I believe this was briefly touched on in #114 ([here exactly](https://github.com/emersion/mako/pull/114#discussion_r257463621)) but the current solution just creates a new notification with brand new id.

vilhalmer:
Hmm. This really seems like the spec is contradicting itself. The description of the parameter earlier in the spec is:

>The optional notification ID that this notification replaces. The server must atomically (ie with no flicker or other visual cues) replace the given notification with this one. This allows clients to effectively modify the notification while it's active. A value of value of 0 means that this notification won't replace any existing notifications.

…which implies that it's only supposed to function if a notification with that id currently exists. This is also complicated by the fact that the spec explicitly endorses using incrementing integer ids, but allowing this can create a situation where the static id you've been using as `replaces_id` happens to be assigned to an unrelated notification.

That said, I'm not particularly opposed to changing this behavior since it makes it easier to interact with the notification, and the case where the ids collide is exactly one notification long.

TheEdward162:
I don't think the spec is contradicting itself as much as just it doesn't really specify what to do in that case. It just all comes down to wording, really, as I would argue that you can "replace" a non-exitstent notification with an existing one.

As for the collision, it could either be ignored, or this case could be handled by keeping a separate HashSet or something similar with replaced ids and skipping those for new notifications. It might be worth looking at how dunst handles it as well.

Would you have time to implement this sometime soon™?

emersion:
>Would you have time to implement this sometime soon™?

No. If you want something to happen, please send patches.

PR #168, opened 2019-06-29
nonexistent replaces_id

TheEdward162:
Referencing #167

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -99,15 +99,18 @@ static int handle_notify(sd_bus_message *msg, void *data,
 	struct mako_notification *notif = NULL;
 	if (replaces_id > 0) {
 		notif = get_notification(state, replaces_id);
-	}
-
-	if (notif) {
-		reset_notification(notif);
+		if (notif) {
+			// Notification with replaces_id was found
+			reset_notification(notif);
+		} else {
+			// Notification with replace_id was not found,
+			// so we create a new one and assign the id to it.
+			notif = create_notification(state, replaces_id);
+			replaces_id = 0; // So we don't attempt to replace it since it doesn't exist yet.
+		}
 	} else {
-		// Either we had no replaces_id, or the id given was invalid. Either
-		// way, make a new notification.
 		replaces_id = 0; // In case they got lucky and passed the next id.
-		notif = create_notification(state);
+		notif = create_notification(state, 0);
 	}
 
 	if (notif == NULL) {
--- a/include/mako.h
+++ b/include/mako.h
@@ -14,6 +14,7 @@
 #include "pool-buffer.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 #include "xdg-output-unstable-v1-client-protocol.h"
+#include "set.h"
 
 struct mako_state {
 	struct mako_config config;
@@ -45,6 +46,7 @@ struct mako_state {
 	struct pool_buffer *current_buffer;
 
 	uint32_t last_id;
+	struct mako_set_top *replaced_ids_set;
 	struct wl_list notifications; // mako_notification::link
 
 	int argc;
--- a/include/notification.h
+++ b/include/notification.h
@@ -73,7 +73,7 @@ typedef char *(*mako_format_func_t)(char variable, bool *markup, void *data);
 bool hotspot_at(struct mako_hotspot *hotspot, int32_t x, int32_t y);
 
 void reset_notification(struct mako_notification *notif);
-struct mako_notification *create_notification(struct mako_state *state);
+struct mako_notification *create_notification(struct mako_state *state, uint32_t replaces_id);
 void destroy_notification(struct mako_notification *notif);
 
 void close_notification(struct mako_notification *notif,
--- a/include/set.h
+++ b/include/set.h
@@ -1,0 +1,32 @@
+#ifndef MAKO_SET_H
+#define MAKO_SET_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+struct mako_set_top {
+	struct mako_set_second *tables[0xFF];
+};
+
+struct mako_set_second {
+	struct mako_set_third *tables[0xFF];
+};
+
+struct mako_set_third {
+	struct mako_set_last *tables[0xFF];
+};
+
+struct mako_set_last {
+	bool values[0xFF];
+};
+
+/// Returns true if the set contains the value.
+bool mako_set_contains(struct mako_set_top *top, uint32_t value);
+/// Inserts the value into the set, allocating tables as needed.
+void mako_set_insert(struct mako_set_top **top, uint32_t value);
+/// Removes the value from the set, returns true if the value existed.
+bool mako_set_remove(struct mako_set_top *top, uint32_t value);
+/// Shrinks the set by deallocating empty tables.
+void mako_set_shrink(struct mako_set_top **top);
+
+#endif
--- a/meson.build
+++ b/meson.build
@@ -61,6 +61,7 @@ src_files = [
 	'types.c',
 	'icon.c',
 	'string-util.c',
+	'set.c',
 ]
 
 if gdk_pixbuf.found()
--- a/notification.c
+++ b/notification.c
@@ -19,6 +19,7 @@
 #include "notification.h"
 #include "icon.h"
 #include "string-util.h"
+#include "set.h"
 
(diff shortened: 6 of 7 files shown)


TheEdward162:
The usecase is that I have a volume notification script that runs `notfiy-send.sh -r 8000 Volume $volume` when the volume changes. I'd like to avoid having to store the returned id in a temporary file, plus the spec doesn't exactly consider this case, so I think this is the best thing to do user experience-wise. Either way, `mako` returns a new ID for non-existent `replaces_id` even if the `replaces_id < new_id`, which is kind of weird. If the user misses the replace notification by just a moment, a different ID is generated, which brings us back to temporary files.

Right now on `dunst` this only ever uses notification with id 8000 for this command, even if one doesn't exist yet. It doesn't, however, remove the replace-created ID from the pool of natural IDs. This seems to work well enough anyway.

> increment the new ID until there's no existing notification with such an ID.

I don't understand what you mean by this. Old notifications (even the replaced ones) are deleted and their IDs are forgotten. I added the primitive hashset to skip IDs which were replace-created in natural ID generation so we don't run into cases where you get a naturally generated notification with the same ID as one that was replace-created previously.

vilhalmer:
I used to feel that the temp files were kind of a hack as well, but if you store them in tmpfs they never really exist anyway.

>Right now on dunst this only ever uses notification with id 8000 for this command, even if one doesn't exist yet. It doesn't, however, remove the replace-created ID from the pool of natural IDs. This seems to work well enough anyway.

If dunst isn't doing anything special about the id, I'd much rather follow its example than add an entire new data structure to store one or two integers.

>increment the new ID until there's no existing notification with such an ID.

I think @emersion means that when we see a new incoming notification without a `replaces_id` (i.e. most of them), iterate through any notifications that happen to exist at the time and make sure the newly assigned id is higher than any of them. This might skip some ids, but we don't actually care that they're contiguous. In your case this would mean when the first notification comes in while the volume display is open, assigned ids will jump from something low to 8001 and then continue counting.

emersion:
We could also go through all notifications and pick the lowest unused ID.

TheEdward162:
Okay, I'll remove the hashset.

> iterate through any notifications that happen to exist at the time and make sure the newly assigned id is higher than any of them.

I think this can easily lead to integer overflow if someone requests MAXINT as `replaces_id`.

> We could also go through all notifications and pick the lowest unused ID.

ID reuse in non-replace notifications is disallowed by the spec, unless overflow happens.

vilhalmer:
>I think this can easily lead to integer overflow if someone requests MAXINT as replaces_id.

The fact that there is so much nuance to this is leading me back to thinking it’s not worth allowing arbitrary ids to be chosen. Missing the notification you’re replacing isn’t really a problem, since you still achieve the goal of not having two on screen at once. Needing to store the id somewhere is not really not that surprising of a requirement given that you’re trying to interact with a stateful daemon, it just happens that `notify-send` has no way of doing it internally since it is short-lived. Allowing an external tool to dictate the value of what is supposed to be an internal tracking mechanism seems like it can only ever lead to bugs or weird edges.

TheEdward162:
As soon as the id is given to the user and encouraged to be used as an identifier I don't think it can be called internal anymore. The tracking wouldn't be needed at all if there was no external `replaces_id`.

If you want to avoid the complexity of the hashset, I would still propose to at least do what `dunst` does and improve the quality of UX even if it can potentially lead to a small window of time where there are two notifications with the same id.

Plus the idea of using the lowest currently unused ID only needs one addition to adhere to the spec - start the search from `new_id` upwards.

emersion:
I agree with @vilhalmer: that the use-case you describe (using a hardcoded ID) seems harmful to support. I don't want to encourage people to do this, the API was never intended to be used this way.

It seems like dunst offers another API for this: `set_stack_tag`. The name is a little misleading.

> The stack tag can be set by the client with the 'synchronous', 'private-synchronous' 'x-canonical-private-synchronous' or the 'x-dunst-stack-tag' hints.

TheEdward162:
Well okay, if you consider improving user experience harmful, I'll close the issue. I don't think we can make any assumptions on how the API was designed to be used since it doesn't fully explain what to do in this case.

emersion:
I think it's pretty clear from the name that you're only supposed to provide `replace_id` if you have a handle to an existing notification.

>Well okay, if you consider improving user experience harmful, I'll close the issue

I'm not against adding another way to improve UX, with something similar to dunst's solution.

TheEdward162:
I'm confused now. Can you please tell me exactly which of the proposed solutions are you least opposed to? 

I'm happy to implement any of them, but I though it would be best to improve on the dunst solution by not allowing natural ID notfications to mix with the replace-created ones. Normally this would be an issue of using the stdlib HashSet, but since the project is in C and I didn't want to add a dependency on something random, I just added a very primitive implementation of a u32 HashSet.

Just to reiterate, the current patch solves all of the discussed issues - at the cost of added complexity. If you wouldn't be opposed to adding a dependency on some collections library (I don't really know C libraries), maybe that could be a better solution?

Lastly, the spec says "If replaces_id is not 0, the returned value is the same value as replaces_id.". This statement is on a separate line and clearly separated from the actual function of `replaces_id`. This leads me to believe that the semantical meaning of the name `replaces_id` really doesn't have much to do with the statement as a separate point. That is, if we were to adhere to this single line as a separate rule, it would mean that even if `replaces_id` was an non-existent id, we would still be obligated to return the same number (even if we chose not to actually display a new notification).

emersion:
What I don't like about returning the replace_id even if the notification doesn't exist is that scripts use it to hardcode a particular ID. This causes issues because each script needs to use a unique ID, and there's no guarantee the replace_id isn't used by a regular notification (if you don't reboot your machine and get chat notifications, I suppose you can get to 8000 pretty quickly).

So an alternate solution is to use a string token. For instance, sending a notification with "brightness" will replace any notification previously submitted with the same string. Dunst calls this "stack tag".

TheEdward162:
With the HashSet, that exact guarantee is upheld.

I agree that hardcoding values might be hard to keep track of, but that's up to the user, not mako as a deamon. (I'd like to use network ports as an analogy).

I don't know about stact tags. Are they part of the specification in any way, or will I need a separate notification sending tool?

emersion:
>With the HashSet, that exact guarantee is upheld.

How so?

>I agree that hardcoding values might be hard to keep track of, but that's up to the user, not mako as a deamon. (I'd like to use network ports as an analogy).

I disagree. Network ports are a poor analogy because (1) all servers are generally always running as daemons (2) when you try to use both at the same time you have a clear error. 

>I don't know about stact tags. Are they part of the specification in any way, or will I need a separate notification sending tool?

It's a hint. If you use `notify-send`, these can be specified with `-h`.

TheEdward162:
Because replace-created IDs are stored in the HashSet and are skipped in the natural ID selection process.

emersion:
Let's say a normal notification has ID 8000, and your script uses a hardcoded 8000 replace_id. Your script will incorrectly replace the notification.

TheEdward162:
That could happen even if mako didn't support this.

emersion:
>That could happen even if mako didn't support this.

Yeah. Which is the reason why I don't want to support this use-case.

>I would argue that that's not as much of a mako problem as a user problem?

The issue here is that you're asking mako to explicitly support broken scripts.

TheEdward162:
No, I'm asking to let the user decide what is broken and what is not, because one person is not a judge of what people find broken.

emersion:
Sorry, this is not how it works. You're free to have a local patch if you want.

However as a maintainer I want to have the freedom to say no to a patch. Consider the maintainer: https://youtu.be/W2AR1owg0ao?t=1120

TheEdward162:
I'm not forcing you to do anything. If you say no then that means no. The PR and issue are closed already.

apprehensions:
#510 Depends on this, Please merge this.

apprehensions:
fnott supports this, why can't mako?

apprehensions:
> The optional notification ID that this notification replaces. **The server must atomically (ie with no flicker or other visual cues) replace the given notification with this one.**

Does mako not go by the specification? it says here the replaceid is exactly for the purpose that this PR intends to create, why was it wrongly rejected?

emersion:
Please avoid spamming pull requests.

mako behaves like the spec says. There is a "tag" mechanism you can use instead for scripts.

PR #169, opened 2019-07-04
notification.c: use correct format arguments

sauyon:
Minor fix to use `%zu` instead of `%d` to format the `size_t` values.

I don't think this will ever actually break, except *maybe* on insane systems where int is small, if somehow enough notifications get sent to overflow things.... but it feels right to fix it anyway.

Tried to actually get this to break on my machine and discovered that `seq 1 1<<33` breaks my shell :P.

Obligatory corporate stuff (LGTM *is* cool, though, I promise :P):
This was found using LGTM.com, there are some other findings [here](https://lgtm.com/projects/g/emersion/mako/alerts/?mode=list) and you can [enable automated PR analysis](https://lgtm.com/projects/g/irtimmer/moonlight-embedded/ci/) to guard against new issues being introduced.

Full disclosure: I work for Semmle, the company behind LGTM.com

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -188,9 +188,9 @@ char *format_hidden_text(char variable, bool *markup, void *data) {
 	struct mako_hidden_format_data *format_data = data;
 	switch (variable) {
 	case 'h':
-		return mako_asprintf("%d", format_data->hidden);
+		return mako_asprintf("%zu", format_data->hidden);
 	case 't':
-		return mako_asprintf("%d", format_data->count);
+		return mako_asprintf("%zu", format_data->count);
 	}
 	return NULL;
 }


sauyon:
Actually, mako just segfaults when you send ~15k notifications, so it's probably impossible to break this.

vilhalmer:
>mako just segfaults aborts when you send ~15k notifications

Interesting. Is that 15k total, or 15k within some period of time? I regularly accidentally hit 600/second due to a notification bug in my chat client at work, so it seems like I must have crossed 15k total at some point.

sauyon:
> > mako just segfaults aborts when you send ~15k notifications
> 
> Interesting. Is that 15k total, or 15k within some period of time? I regularly accidentally hit 600/second due to a notification bug in my chat client at work, so it seems like I must have crossed 15k total at some point.

Doing a bit more testing suggests that it's a lot of messages in a short time. I managed to get to varying numbers, but the largest I've seen is ~23k so far (with a test program that does nothing but spam notifications).

vilhalmer:
Tracking that down will be a fun rainy day project. :)

Anyway, the PR …lgtm.

vilhalmer:
Thanks!

PR #170, opened 2019-07-04
Initial touch support

### LINKED ISSUE #99: Touch support
would be nice if touch gestures could be supported, e.g. swiping to removing the notification.

sheenobu:
Listen for touch down, motion and up events.

Nind 'press' event to 'up' action. bind 'down' and 'motion'
to x and y tracking.

Implement the button press dispatching as a simple
'mocked' button press using BTN_TOUCH so it fits in
with the existing button press event structure.

Fixes #99

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -7,11 +7,11 @@
 #include "types.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 
-enum mako_button_binding {
-	MAKO_BUTTON_BINDING_NONE,
-	MAKO_BUTTON_BINDING_DISMISS,
-	MAKO_BUTTON_BINDING_DISMISS_ALL,
-	MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION,
+enum mako_binding {
+	MAKO_BINDING_NONE,
+	MAKO_BINDING_DISMISS,
+	MAKO_BINDING_DISMISS_ALL,
+	MAKO_BINDING_INVOKE_DEFAULT_ACTION,
 };
 
 enum mako_sort_criteria {
@@ -81,8 +81,10 @@ struct mako_config {
 	struct mako_style superstyle;
 
 	struct {
-		enum mako_button_binding left, right, middle;
+		enum mako_binding left, right, middle;
 	} button_bindings;
+
+	enum mako_binding touch;
 };
 
 void init_default_config(struct mako_config *config);
--- a/notification.c
+++ b/notification.c
@@ -269,7 +269,7 @@ size_t format_text(const char *format, char *buf, mako_format_func_t format_func
 	return len;
 }
 
-static enum mako_button_binding get_button_binding(struct mako_config *config,
+static enum mako_binding get_binding(struct mako_config *config,
 		uint32_t button) {
 	switch (button) {
 	case BTN_LEFT:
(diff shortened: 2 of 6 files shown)


include/config.h
```diff
@@ -14,6 +14,13 @@ enum mako_button_binding {
 	MAKO_BUTTON_BINDING_INVOKE_DEFAULT_ACTION,
 };
 
+enum mako_touch_binding {
```
emersion: I wonder if it would make sense to unify this enum with `mako_button_binding`

notification.c
```diff
@@ -310,6 +310,33 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
 	}
 }
 
+static enum mako_touch_binding get_touch_binding(struct mako_config *config) {
```
emersion: No need for this function, can use `config->touch_bindings.press` directly

notification.c
```diff
@@ -310,6 +310,33 @@ void notification_handle_button(struct mako_notification *notif, uint32_t button
...
+	return config->touch_bindings.press;
+}
+
+void notification_handle_touch(struct mako_notification *notif) {
```
emersion: Yeah, this duplicates the code. Probably a good idea to unify button and touch handling.

wayland.c
```diff
@@ -129,6 +152,15 @@ static const struct wl_pointer_listener pointer_listener = {
 	.axis = noop,
 };
 
+static const struct wl_touch_listener touch_listener = {
+	.down = noop,
```
emersion: It's possible to get a down event without a motion event if you press and release without moving. The down event has x, y so we should handle it and update our state.

wayland.c
```diff
@@ -96,6 +96,29 @@ static void destroy_output(struct mako_output *output) {
...
+static void touch_handle_motion(void *data, struct wl_touch *wl_touch,
+		uint32_t time, int32_t id,
+		wl_fixed_t surface_x, wl_fixed_t surface_y) {
+	struct mako_seat *seat = data;
+	seat->touch.x = wl_fixed_to_int(surface_x);
+	seat->touch.y = wl_fixed_to_int(surface_y);
+}
+
+static void touch_handle_up(void *data, struct wl_touch *wl_touch,
+		uint32_t serial, uint32_t time, int32_t id) {
```
emersion: Follow-up issue: handle `id` to add support for multitouch.

wayland.c
```diff
@@ -143,6 +175,11 @@ static void seat_handle_capabilities(void *data, struct wl_seat *wl_seat,
 		wl_pointer_add_listener(seat->pointer.wl_pointer,
 			&pointer_listener, seat);
 	}
+	if (capabilities & WL_SEAT_CAPABILITY_TOUCH) {
+		seat->touch.wl_touch = wl_seat_get_touch(wl_seat);
```
emersion: In case `seat->touch.wl_touch` is already set, we need to release it. See the first `if` block.

wayland.c
sheenobu: that makes sense!

notification.c
sheenobu: I introduced this for the potential future when mako has more than just one touch_binding (single-touch, double-touch, gestures). it's not necessary though, so I'll follow your suggestion.

sheenobu: made press as part of get_button_binding.

wayland.c
```diff
@@ -96,6 +101,40 @@ static void destroy_output(struct mako_output *output) {
...
+	struct mako_state *state = seat->state;
+
+	struct mako_notification *notif;
+	wl_list_for_each(notif, &state->notifications, link) {
+		if (hotspot_at(&notif->hotspot, seat->touch.x, seat->touch.y)) {
+			// simulate a touch as a simple button press of type
+			// BTN_TOUCH
+			notification_handle_button(notif, BTN_TOUCH,
+				WL_POINTER_BUTTON_STATE_PRESSED);
+			break;
```
sheenobu: i don't love this.

emersion: Indeed. What about extracting most of `notification_handle_button`'s code into a new `notification_execute_binding(struct mako_notification *notif, enum mako_button_binding binding)` function?

sheenobu: i think that works, updated! thanks.

include/config.h
```diff
@@ -81,7 +81,7 @@ struct mako_config {
 	struct mako_style superstyle;
 
 	struct {
-		enum mako_button_binding left, right, middle;
+		enum mako_button_binding left, right, middle, touch;
```
emersion: Can we move the new `touch` member out of the `button_bindings` struct? It's not a button binding.

sheenobu: no worries, it's still considered a type of mako_button_binding ?

emersion: What about renaming `mako_button_binding` to `mako_binding` for consistency?

sheenobu: sure

sheenobu:
that was a load-bearing semicolon, re-added

```
diff --git a/notification.c b/notification.c
index 548b2be..3234a43 100644
--- a/notification.c
+++ b/notification.c
@@ -293,7 +293,7 @@ void notification_execute_binding(struct mako_notification *notif,
 	case MAKO_BINDING_DISMISS_ALL:
 		close_all_notifications(notif->state, MAKO_NOTIFICATION_CLOSE_DISMISSED);
 		break;
-	case MAKO_BINDING_INVOKE_DEFAULT_ACTION:
+	case MAKO_BINDING_INVOKE_DEFAULT_ACTION:;
 		struct mako_action *action;
 		wl_list_for_each(action, &notif->actions, link) {
 			if (strcmp(action->key, DEFAULT_ACTION_KEY) == 0) {
```

PR #181, opened 2019-07-24
Include list command in makoctl zsh completions

RPigott:


contrib/completions/zsh/_makoctl
```diff
@@ -5,6 +5,7 @@ local -a makoctl_cmds
 makoctl_cmds=(
     'dismiss:Dismisses notification (first by default)'
     'invoke:Invokes an action on the first notification. If action is not specified, invokes the default action'
+	'list:Retrieve a list of current notifications'
```
emersion: Hmm, tab vs space issue. Can you convert the whole file to tabs?

RPigott: Okay, I changed it to tabs.

PR #182, opened 2019-07-25
docs: Split out configuration file docs

mcoffin:
Split the configuration file documentation out from mako.1 into
mako.5. This will help users find the configuration file man page
without having to scroll so deeply into mako(1).

### DIFF
--- a/mako.1.scd
+++ b/mako.1.scd
@@ -183,96 +183,15 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: none
 
-# CRITERIA-ONLY STYLE OPTIONS
+# CONFIGURATION FILE
 
-Some style options are not useful in the global context and therefore have no
-associated command-line option.
+The mako configuration file is located at *~/.config/mako/config* or at
+*$XDG\_CONFIG\_HOME/mako/config*.
 
-*invisible* 0|1
-	Whether this notification should be invisible even if it is above the
-	_max-visible_ cutoff. This is used primarily for hiding members of groups.
-	If you want to make more than the first group member visible, turn this
-	option off within a _group-index_ criteria.
+The configuration file supports *CRITERIA* with allows changing behavior
+per-notifiaction, per-application, or on other conditions.
 
-	Default: 0
-
-# CONFIG FILE
-
-The config file is located at *~/.config/mako/config* or at
-*$XDG\_CONFIG\_HOME/mako/config*. Each line of the form:
-
-	key=value
-
-Is equivalent to passing *--key=value* to mako from the command line. Note that
-any quotes used within your shell are unnecessary and also invalid in the
-config file.
-
-Empty lines and lines that begin with # are ignored.
-
-# CRITERIA
-
-In addition to the set of options at the top of the file, the config file may
-contain zero or more sections, each containing any combination of the
-*STYLE OPTIONS*. The sections, called criteria, are defined with an INI-like
-square bracket syntax. The brackets may contain any number of fields, like so:
-
-	\[field=value field2=value2 ...\]
-
-When a notification is received, it will be compared to the fields defined in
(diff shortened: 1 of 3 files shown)


mako.1.scd
```diff
@@ -183,96 +183,15 @@ dismissed with a click or via *makoctl*(1).
...
+The mako configuration file is located at *~/.config/mako/config* or at
+*$XDG\_CONFIG\_HOME/mako/config*.
 
-*invisible* 0|1
-	Whether this notification should be invisible even if it is above the
-	_max-visible_ cutoff. This is used primarily for hiding members of groups.
-	If you want to make more than the first group member visible, turn this
-	option off within a _group-index_ criteria.
+The configuration file supports *CRITERIA* with allows changing behavior
+per-notifiaction, per-application, or on other conditions.
```
vilhalmer: Typos: "with allows", "notifiaction".

This makes it sound like you can only use criteria in the config file, and not also the globals. How about something like "in addition to the same configuration supplied by the command line flags, the configuration file also allows overriding styles based on the source and content of each notification".

mako.1.scd
```diff
@@ -183,96 +183,15 @@ dismissed with a click or via *makoctl*(1).
 
 	Default: none
 
-# CRITERIA-ONLY STYLE OPTIONS
+# CONFIGURATION FILE
 
-Some style options are not useful in the global context and therefore have no
-associated command-line option.
+The mako configuration file is located at *~/.config/mako/config* or at
+*$XDG\_CONFIG\_HOME/mako/config*.
```
vilhalmer: I would leave this out since it is also present in mako.5, so that we don't end up forgetting to update one or the other at some point down the road.

mako.5.scd
```diff
@@ -0,0 +1,118 @@
...
+
+# DESCRIPTION
+
+The mako configuration file is located at *~/.config/mako/config* or at
+*$XDG\_CONFIG\_HOME/mako/config*. Each line of the form:
+
+	key=value
+
+Is equivalent to passing *--key=value* to mako from the command line. See
+*mako*(1) for a list of command line options. Note that any quotes used within
```
vilhalmer: This bouncing back and forth is why I'm slightly hesitant to break these up. Would it make more sense to just alter the flow of the manpage so that users learn they can use the config file before seeing all of the flags? That section on its own is very short and could be right under the help flag section.

emersion:
I think the main way to configure mako is via the config file, so maybe it makes sense to move all options to `mako(5)`.

mcoffin:
I agree. I’ll make the changes to move that section over tonight. :)

mcoffin:
@vilhalmer FYI I chose this approach

1. Becuase I assumed that it didn’t have a config file due to lack of a man5 page.
2. That seemed like the best path forward after a short IRC back and forth with @emersion.

If you’ve got some suggestions on a better path forward (like simply duplicating the options sections), obviously I’m open to it.

Just trying to improve usability for new people since I encountered issues when transitioning from dunst.

vilhalmer:
I’m on board with moving the whole thing into mako.5. :)

vilhalmer:
Gentle ping. :)

mcoffin:
Sorry. Yes. I will update this probably tomorrow. Normally, I'm pretty responsive on GitHub and whatnot, but I've been working on some kernel stuff, and honestly this went totally by the wayside, but I'll finish it up tomorrow haha. Sorry for dirtying up your PR queue for so long. Totally my bad for falling off the face of the planet.

vilhalmer:
No worries, we're never in a hurry. :)

emersion:
@mcoffin Hi again, please let me know if you prefer to let someone else take over this PR :)

emersion:
Superseded by https://github.com/emersion/mako/pull/216

Thanks for taking the time to work on this anyway!

Issue #171, opened 2019-07-05
Add --config option


Dan-Theriault:
It would be useful for the CLI to have an option to set the path to the configuration.

emersion:
Sure. Patches welcome :)

emersion:
Fixed in https://github.com/emersion/mako/pull/190

PR #190, opened 2019-10-03
Add --config option

dg10a:
Addresses https://github.com/emersion/mako/issues/171

Not sure if there are issues with looping over the arguments twice. Feedback welcome.

### DIFF
--- a/config.c
+++ b/config.c
@@ -544,10 +544,15 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-int load_config_file(struct mako_config *config) {
-	char *path = get_config_path();
-	if (!path) {
-		return 0;
+int load_config_file(struct mako_config *config, char *config_arg) {
+	char *path = NULL;
+	if (config_arg == NULL) {
+		path = get_config_path();
+		if (!path) {
+			return 0;
+		}
+	} else {
+		path = config_arg;
 	}
 
 	FILE *f = fopen(path, "r");
@@ -642,6 +647,7 @@ int load_config_file(struct mako_config *config) {
 int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 	static const struct option long_options[] = {
 		{"help", no_argument, 0, 'h'},
+		{"config", required_argument, 0, 'c'},
 		{"font", required_argument, 0, 0},
 		{"background-color", required_argument, 0, 0},
 		{"text-color", required_argument, 0, 0},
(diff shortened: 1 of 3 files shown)


emersion:
Ah, interesting, someone sent me a very similar patch some time ago: https://lists.sr.ht/~emersion/public-inbox/patches/8076

dg10a:
Glad to see it's being worked on. Let me know if you want me to close this PR in favor of that patch.

emersion:
I'd be interested to see an approach that addresses the concerns explained there: https://lists.sr.ht/~emersion/public-inbox/patches/8076#%3CtpSpmtft2kY55ktdn17GJ7UCRZHWb7Qxpr97WfHwMEcVL51_ighZLa50bdYLZcsMCn1sRHXKFJvpfpsqj4qy-9svQpg6V3snyDNB1ZhvAKc=@emersion.fr%3E

dg10a:
I think the current PR meets your concerns, but it loops with `getopt` twice. What do you think of the approach below? It works neatly for the style options, but I'm not sure how to handle the config options.

```C
// store style options
struct mako_style opt_style;
init_empty_style(&opt_style);

optind = 1;
char *config_arg = NULL;
while (1) {
	int option_index = -1;
	int c = getopt_long(argc, argv, "hc:", long_options, &option_index);
	if (c < 0) {
		break;
	} else if (c == 'h') {
		return 1;
	} else if (c == 'c') {
		config_arg = strdup(optarg);
	} else if (c != 0) {
		return -1;
	} else {
		const char *name = long_options[option_index].name;
		if (!apply_style_option(&opt_style, name, optarg)
				&& !apply_config_option(config, name, optarg)) {
			fprintf(stderr, "Failed to parse option '%s'\n", name);
			return -1;
		}
	}
}

int config_status = load_config_file(config, config_arg);
if (config_status < 0) {
	return -1;
}

// apply style options
apply_style(&root_criteria->style, &opt_style);
finish_style(&opt_style);
```

config.c
```diff
@@ -544,10 +544,15 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-int load_config_file(struct mako_config *config) {
-	char *path = get_config_path();
-	if (!path) {
-		return 0;
+int load_config_file(struct mako_config *config, char* config_arg) {
```
emersion: Style: `*` on the right

PR #194, opened 2019-10-08
Support choosing a notification action with `makoctl menu`

### LINKED ISSUE #4: Support choosing a notification action
Do we want to use dmenu for this?

xPMo:
This is a mostly-straightforward inclusion of [my menu selection script](https://gist.github.com/xPMo/462936ca0e86f3697a6365b2be02525d), with associated docs. This would close #4. 

I include it under `makoctl menu`, feel free to suggest alternate subcommand names, or any other criticism. I removed or reduced most comments as well, although I'm tempted to include the longer explanations from the gist in the body of the commit message for the sake of `git blame`.

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -14,7 +14,7 @@ usage() {
 }
 
 call() {
-	busctl -j --user call org.freedesktop.Notifications /fr/emersion/Mako \
+	busctl --user call org.freedesktop.Notifications /fr/emersion/Mako \
 		fr.emersion.Mako "$@"
 }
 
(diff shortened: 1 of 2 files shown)


emersion:
Hmm, can we use `jq` instead of all of this string manipulation?

xPMo:
We could also reduce some complexity by changing from POSIX sh to Bash. Then we could use native `${parameter//substitutions}` instead of sed, and use associative arrays instead of abusing the argument list with `set --`. I would call this a middle ground, but it's still not great compared to a `jq` script.

xPMo:
[jq-based version](https://github.com/xPMo/mako/tree/jqmenu), integrated into makoctl just like this one.

emersion:
To be honest, I'd rather depend on `jq` rather than doing string manipulations. Another option would be to rewrite `makoctl` in C.

xPMo:
Closing in favor of #195.

PR #195, opened 2019-10-10
Support choosing a notification action with 'makoctl menu'

### LINKED ISSUE #4: Support choosing a notification action

xPMo:
Requires `jq`, alternative to #194. Would close #4.

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -8,6 +8,8 @@ usage() {
 	echo "  invoke [-n id] [action]  Invoke an action on the notification"
 	echo "                           with the given id, or the last"
 	echo "                           notification if none is given"
+	echo "  menu [prog] [arg ...]    Use [prog] [args ...] to select one"
+	echo "                           notification action to be invoked"
 	echo "  list                     List notifications"
 	echo "  reload                   Reload the configuration file"
 	echo "  help                     Show this help"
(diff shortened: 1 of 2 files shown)


makoctl
```diff
@@ -53,6 +55,30 @@ case "$1" in
...
+	shift 1
+	if ! type -p jq > /dev/null; then
+		echo >&2 "$0: jq is required to use 'menu'"
+		exit 1
+	fi
+	if array="$(call ListNotifications | jq -re '.data[0][0].actions.data | if length > 0 then . else false end')"; then
+		sel="$(jq -rn "$array|values[]" | "$@")"
+		sel="$(jq -rn --arg sel "$sel" "$array|to_entries[]|select(.value == \$sel).key")"
+		if [ -n "$sel" ]; then
+			# The key may contain C escape sequences, use printf %b to evaluate them.
```
emersion: Why would the key contain escape sequences that need to be evaluated?

xPMo: It won't, my mistake. We will still need to add a trailing character to prevent newlines from being mangled.

emersion: Honestly it's probably not worth it. Action identifiers shouldn't contain newlines.

xPMo: Fair.

xPMo: Yeah, the odds of encountering it is ridiculously small. I am inclined to guard against a possible future issue, but I'll concede.

xPMo:
Slight tangent; it looks like Drew maintains `mako-git` on the AUR. Could you ask him to add `optdepends=('jq: support for makoctl menu')` to the PKGBUILD?

emersion:
cc @ddevault ^

xPMo:
[Actually, I went ahead and took the opportunity to use `git send-email`](https://lists.sr.ht/~sircmpwn/public-inbox/%3C20191027190435.141513-1-GammaFunction%40vivaldi.net%3E)

p00f:
@emersion Is there supposed to be a "dismiss" action in the wofi menu or am I understanding incorrectly? I only see "activate"

emersion:
No, dismiss is not an action. It's a separate makoctl command.

PR #197, opened 2019-10-21
Add support for tracking multiple touches (fixes #175)

### LINKED ISSUE #175: Add multi-touch support
We have basic touch support, however we don't support multi-touch yet. We need to track multiple touch points for this.

See https://github.com/emersion/mako/pull/170#discussion_r300548507

rianadon:
To keep the code simple I used an array with max capacity 10 for all touchpoints since I couldn't think of any simple way to implement a linked list or map.

Hopefully it's reasonable enough to assume people aren't dismissing notifications with more than 10 fingers (or even more than 5).

It's now anyhow possible to swipe away all your notifications with five carefully placed fingers :raised_hand_with_fingers_splayed:.

Fixes #175.

### DIFF
--- a/wayland.c
+++ b/wayland.c
@@ -100,16 +100,18 @@ static void touch_handle_motion(void *data, struct wl_touch *wl_touch,
 		uint32_t time, int32_t id,
 		wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct mako_seat *seat = data;
-	seat->touch.x = wl_fixed_to_int(surface_x);
-	seat->touch.y = wl_fixed_to_int(surface_y);
+	if (id >= MAX_TOUCHPOINTS) return;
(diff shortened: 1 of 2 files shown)


wayland.c
```diff
@@ -100,16 +100,18 @@ static void touch_handle_motion(void *data, struct wl_touch *wl_touch,
 		uint32_t time, int32_t id,
 		wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct mako_seat *seat = data;
-	seat->touch.x = wl_fixed_to_int(surface_x);
-	seat->touch.y = wl_fixed_to_int(surface_y);
+	if (id >= MAX_TOUCHPOINTS) return;
```
vilhalmer: Style nit: This should be broken out to a block with braces in all three spots.

rianadon: Done!

PR #205, opened 2019-12-11
README: Added package dependencies for Debian

Zocker1999NET:
The idea is to simplify the build process for users of certain distros so they do not search on their own for the right packages.

For example, `meson` complained about missing the `wayland-client` library, so I have installed every package having `wayland-client` in their name, but the required package was `libwayland-dev`.
Similar with `pango`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ Install dependencies:
 * scdoc (optional, for man pages)
 * jq (optional, runtime dependency)
 
+Required packages for the dependencies on Debian (Unstable):
+
+* `meson`
+* `libwayland-dev`
+* `wayland-protocols`
+* `libcairo2-pango-ocaml-dev`
+* `libcairo2-dev`
+* `libsystemd-dev` or `libelogind-dev`
+* `libgdk-pixbuf2.0-dev`
+* `scdoc` (optional, for man pages)
+* `jq`(optional, runtime dependency)
+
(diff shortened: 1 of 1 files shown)


emersion:
Sorry, I don't want to add distro-specific instructions to the README.

markstos:
Could this live on a Github wiki page where it could be maintained by mako's Debian user base?

I'm working on trying to make sway-gnome easier to install. It depends on mako, so Debian and Ubuntu users of sway-gnome are also interested the details of what's involved in getting mako installed in that environment. 

Currently the wiki feature is disabled on this repo.

markstos:
Related, there is an "intent to package" `mako` as `mako-wayland` for Debian, so it will be appearing in Debian and Ubuntu in due time.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=912865

markstos:
More about that release is here:  https://ftp-master.debian.org/new/mako-notifier_1.4-1.html

Their git repo is here:  https://salsa.debian.org/swaywm-team/mako (This adds the Debian packaging details)

emersion:
Sure, just enabled the wiki,. Feel free to create a Debian page.

markstos:
Thanks. I'll note that `mako-notifier` is now in Debian Unstable: https://packages.debian.org/sid/mako-notifier

PR #216, opened 2020-01-11
Split out configuration docs

emersion:
Create a new mako(5) man page for the configuration file. Configuration
options are moved there since most people will use the config file to
set them.

Supersedes: https://github.com/emersion/mako/pull/182


vilhalmer:
LGTM, thanks!

Issue #138, opened 2019-04-11
Proposal: allow to change configuration via IPC


emersion:
We could add a makoctl command like this:

```
makoctl set [app-name="whatever"] border-color=red
```

These configuration options would be added to the current configuration. They would stay ephemeral, ie. `makoctl reload` would unset those.

The idea is to allow for things like:

- Temporarily hiding all notifications (ie. when doing a presentation or showing a screen to a third-party)
- Change the theme during the day
- Probably other things too

This would be similar to `swaymsg <command>`.

Thoughts?

vilhalmer:
I like this idea, and it should be fairly easy to add since it would just involve inserting another criteria at the end of the list.

My main concern is that a long-running instance will start to collect a large number of criteria entries. This is especially true if something like this were used by scripting such for your time-of-day example. We haven't really explored the results of allowing them to build up.

A secondary concern is that some of the style options 'stack' in ways that can be somewhat non-intuitive, and would be difficult to cancel out without reloading the configuration and potentially losing other state that you would rather have kept.

I don't think either of these should block adding this though, as both can always be improved later on if they prove to be real problems. Something that could potentially fix/avoid both at the same time is adding some sort of user-defined name to criteria, which would then cause the style of that one to be changed when referenced rather than inserting a new one.

CameronNemo:
Rather than going through dbus, it may be more natural to read any available configs in, for example `XDG_RUNTIME_DIR/mako/config.d`.

emersion:
>My main concern is that a long-running instance will start to collect a large number of criteria entries. 

We can compare criteria, and merge the new one with an existing one if there is any.

>A secondary concern is that some of the style options 'stack' in ways that can be somewhat non-intuitive, and would be difficult to cancel out without reloading the configuration and potentially losing other state that you would rather have kept.

Hmm, do you have an example of this?

>Rather than going through dbus, it may be more natural to read any available configs in, for example $XDG_RUNTIME_DIR/mako/config.d.

Seems it would be more work for us and for the user.

vilhalmer:
>Hmm, do you have an example of this?

Non-intuitive might be the wrong term. "Inconvenient" is probably more accurate. The order of the config file is important for things to behave the way you want, and adding new criteria at runtime may not interact with this very well.

Using the following hypothetical config file:
```
[app-name=Firefox]
background-color=#cccccc

[actionable]
background-color=#dd0000
```
[...]

PR #218, opened 2020-01-12
Add `makoctl set`

emersion:
This can be used for instance to implement a "do not disturb" mode:

    makoctl set invisible=1
    makoctl set invisible=0

This is similar to `swaymsg <command>`.

Only global configuration options are supported for now. Criteria requires more discussion.

References: https://github.com/emersion/mako/issues/138

### DIFF
--- a/config.c
+++ b/config.c
@@ -508,6 +508,13 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	return false;
 }
 
+bool apply_global_option(struct mako_config *config, const char *name,
+		const char *value) {
+	struct mako_criteria *global = global_criteria(config);
+	return apply_style_option(&global->style, name, value) ||
+		apply_config_option(config, name, value);
+}
+
 static bool file_exists(const char *path) {
 	return path && access(path, R_OK) != -1;
 }
@@ -676,9 +683,6 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{0},
 	};
 
-	struct mako_criteria *root_criteria =
-		wl_container_of(config->criteria.next, root_criteria, link);
-
 	optind = 1;
 	char *config_arg = NULL;
 	while (1) {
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -508,6 +508,13 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	return false;
 }
 
+bool apply_global_option(struct mako_config *config, const char *name,
+		const char *value) {
+	struct mako_criteria *global = global_criteria(config);
```
vilhalmer: Applying these to the global (first) criteria means you'll only be able to override style options that aren't set by anything in the config file. I'm not sure this is what most users would expect. For example, the do not disturb use case won't work with my config because I make "clever" use of invisible and group-index.

On the other hand, if we inserted a criteria last in the order for IPC to update, `invisible=false` over IPC would break hiding grouped notifications for the default config.

All that is to say, I think updating the global criteria is the right choice for now. Just wanted to get this out there to think about. Not sure what the right solution is. (For do not disturb specifically, we may just want a new config option.)

RobinOpletal:
Is there a possibility to toggle as well or should I make a script for that?

emersion:
No, there's no way to toggle. Use a script.

RobinOpletal:
Thanks :)

fourstepper:
Hi again @emersion - perhaps a bad question, but is there a way to figure out the state of mako set by

```
makoctl set invisible=1
makoctl set invisible=0
```
?

Thanks!

fauust:
Exactly what I was looking for!

@fourstepper I guess you could implement something like this:

```bash
typeset -r VAR_NOTIFY_STATUS="$HOME/.notifications_status"
if grep -q OFF "$VAR_NOTIFY_STATUS"; then
  makoctl set invisible=0
  notify-send -t 1000 -u low "notifications on"
  echo ON >"$VAR_NOTIFY_STATUS"
else
  echo OFF >"$VAR_NOTIFY_STATUS"
  notify-send -t 1000 -u low "notifications off"
  sleep 1 && makoctl set invisible=1
fi
```

PR #220, opened 2020-01-23
Add a history buffer for expired notifications

jktr:
Notifications that have reached their timeout may
be added to a history buffer, displacing the
oldest stored notification if the buffer is full.

The buffer's size is controlled by the global
`max-history` setting. The `history` style
determines if a notification is saved to the
history buffer when it expires.

A new `RestoreNotification` dbus method restores
the most recent notification from the history buffer.

Restored notifications must be dismissed manually.

Closes: https://github.com/emersion/mako/issues/91

### DIFF
--- a/contrib/completions/zsh/_mako
+++ b/contrib/completions/zsh/_mako
@@ -20,7 +20,9 @@ _arguments \
     '--actions[Applications may request an action to be associated with activating a notification. Disabling this will cause mako to ignore these requests.]:action enabled:(0 1)' \
     '--format[Format string.]:format:' \
     '--hidden-format[Format string.]:format:' \
-    '--max-visible[Max number of visible notifications.]:visible notification:' \
+    '--max-visible[Max number of visible notifications.]:visible notifications:' \
+    '--max-history[Max size of history buffer.]:historical notifications:' \
+    '--history[Add expired notification to history.]:history:' \
     '--default-timeout[Default timeout in milliseconds.]:timeout (ms):' \
     '--ignore-timeout[If set, mako will ignore the expire timeout sent by notifications and use the one provided by default-timeout instead.]:Use default timeout:(0 1)' \
     '--output[Show notifications on this output.]:name:' \
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -78,6 +78,25 @@ done:
 	return sd_bus_reply_method_return(msg, "");
 }
 
+static int handle_restore_action(sd_bus_message *msg, void *data,
+		sd_bus_error *ret_error) {
+	struct mako_state *state = data;
+
+	if (wl_list_empty(&state->history)) {
+		goto done;
+	}
+
+	struct mako_notification *notif =
+		wl_container_of(state->history.next, notif, link);
+	wl_list_remove(&notif->link);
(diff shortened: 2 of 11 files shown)


jktr:
This is derived from the discussion in #91, but currently provides only a minimal, but functional implementation. 

Discussed features that are notably missing are
1. a `makoctl history` command for querying history
2. a non-"sticky" history option/style — restored notifications must currently be expired manually
3. a restore-N option for `makoctl restore`

jktr:
The `history` style should apply correctly as of 11c69df, and notification's timers are now cancelled before appending them to the history buffer.

notification.c
```diff
@@ -106,7 +106,22 @@ void close_notification(struct mako_notification *notif,
 		free(notif_criteria);
 	}
 
-	destroy_notification(notif);
+	if (!notif->style.history ||
+	    notif->state->config.max_history <= 0 ) {
```
emersion: Style: we use two tabs to indent `if` continuation lines

notification.c
```diff
@@ -106,7 +106,22 @@ void close_notification(struct mako_notification *notif,
...
+		destroy_notification(notif);
+		return;
+	}
+
+	destroy_timer(notif->timer);
+	notif->timer = NULL;
+
+	wl_list_insert(&notif->state->history, &notif->link);
+	while (wl_list_length(&notif->state->history) >
+	       notif->state->config.max_history) {
```
emersion: Ditto

dbus/mako.c
```diff
@@ -78,6 +78,26 @@ static int handle_invoke_action(sd_bus_message *msg, void *data,
...
+	struct mako_state *state = data;
+
+	if (wl_list_empty(&state->history)) {
+		goto done;
+	}
+	
+	struct mako_notification *notif =
+		wl_container_of(state->history.next, notif, link);
+	wl_list_remove(&notif->link);
+	wl_list_init(&notif->link);
```
emersion: No need to init the link here, `insert_notification` will do it.

mako.5.scd
```diff
@@ -150,6 +157,13 @@ Empty lines and lines that begin with # are ignored.
 
 	Default: 1
 
+*history* = 0|1
```
emersion: Making this allows to disable history selectively for some notifications with criteria, nice.

notification.c
emersion: Style: no space before `)`

jktr:
Thanks for the feedback! The issues you mentioned should be fixed now.

In addition:
* `usage` in `main.c` has been amended
* `finish` in `main.c` now also cleans up the notifications in the history buffer
* zsh completions in `contrib/completions/zsh/` have been amended

contrib/completions/zsh/_mako
```diff
@@ -21,6 +21,8 @@ _arguments \
     '--format[Format string.]:format:' \
     '--hidden-format[Format string.]:format:' \
     '--max-visible[Max number of visible notifications.]:visible notification:' \
+    '--max-history[Max size of history buffer.]:historical notifictions:' \
```
emersion: Typo: notification

PR #223, opened 2020-01-23
Implement `do-not-disturb` via `max_visible=0`

maximbaz:
Dunst has a very nice feature of being able to temporarily silence all notifications. I'd like to be able to have the same "do not disturb" mode with mako as well.

I saw you implemented `makoctl set` command to temporarily change config values, but it seems there is no config to stop or start mako (unless I'm missing something obvious)?

Continuing the logic, instead of implementing a completely new config, I was thinking we could make use of the existing `max-visible` setting?

In this case, `makoctl set max-visible=0` will enable "do not disturb" mode, and `makoctl set max-visible=5` will disable it.

I now treat `max-visible=0` specially to not show `(1 more)` message, but otherwise the patch proved to be quite simple.

What do you think, about the idea, and the code in general?

### DIFF
--- a/render.c
+++ b/render.c
@@ -258,6 +258,11 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 	int pending_bottom_margin = 0;
 	struct mako_notification *notif;
 	wl_list_for_each(notif, &state->notifications, link) {
+		if (config->max_visible >= 0 &&
+				visible_count >= (size_t)config->max_visible) {
+			break;
+		}
+
 		// Note that by this point, everything in the style is guaranteed to
 		// be specified, so we don't need to check.
 		struct mako_style *style = &notif->style;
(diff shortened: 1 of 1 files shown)


emersion:
>it seems there is no config to stop or start mako

Hmm, the user is responsible for starting and stopping mako.

>Continuing the logic, instead of implementing a completely new config, I was thinking we could make use of the existing max-visible setting?

What about just using the `invisible` option?

```
makoctl set invisible=1
makoctl set invisible=0
```

cyrinux:
Hi @emersion, thanks for your work.
Can you implement `GetConfigOption` and `get` in makoctl?
Regards

emersion:
Why is this necessary? The script calling `set` can keep some state.

maximbaz:
Nice, thanks for the hint with `invisible=1`, I somehow missed that this exists!

PR #227, opened 2020-01-30
Remove the bottom margin for the last notification

### LINKED ISSUE #225: Border rendering with multiple notifications (regression ?)
2020-01-29-180525_screenshot
See the picture (before on the right, current on the left) with the exact same config.

For multiple notifications, the border size was accounted once (border were overlapping ?). Is the new behavior what is expected ? Is there a way to change the behavior to the previous one (negative margin ?) ?

tezoi:
If the notification is the last one, the margin is set to `0`.
This is minimal and complement f8ddfe048072d85e1d78cb486a56634b1b93e200 in response to #225, (and potentially allows a nice notification ovelap too). The old behavior is restored by setting negative top and bottom margins values to the border width value.

Overlap example: `margin=-10,0,-10,0` 
2020-01-30-160623_screenshot

Fixes #225

### DIFF
--- a/render.c
+++ b/render.c
@@ -310,9 +310,9 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 
 		total_height += notif_height;
 		pending_bottom_margin = style->margin.bottom;
-                if (i == count){
-	            pending_bottom_margin = 0;
-                }
+		if (i == count){
+			pending_bottom_margin = 0;
+		}
 
 		if (notif->group_index < 1) {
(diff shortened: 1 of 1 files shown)


render.c
```diff
@@ -309,6 +310,9 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 
 		total_height += notif_height;
 		pending_bottom_margin = style->margin.bottom;
+                if (i == count){
```
emersion: Style: we use tabs

Issue #123, opened 2019-03-16
Anchor is not a style parameter and cannot be redefined in criteria section


TLogiviere:
Hello, thank you for making this software!
It would be great to allow notifications to be displayed in different places depending on the application generating them.
For example, one could very much like to make sound and brightness notifications to display at bottom-centre ; chat at centre-left ; system info at top-right and e-mails at top-centre.
Moreover it would further reduce the risk that a given notification gets cleared out by others appearing at the same time.
What do you think about it?

emersion:
I'm not a big fan of this. It becomes unclear what the last notification is, and would make the code a lot more complex.

disruptek:
An alternative might be to run a separate mako instance for each area of the screen on which you wish to see notifications.  Does that sound easier to implement?

I worked around the issue by using varied offsets, but I wished that most of the "global" options could be applied differently according to application.

vilhalmer:
It's impossible to run more than one D-Bus notification client, only one can have the name on the  bus at a given time.

disruptek:
So, harder, then.  :wink:

TLogiviere:
@emersion : Ah. Let me add that the principle of the last (or more depending on `max visible`) notification being displayed would not change, but there would be as many as there is criteria section redefining `--anchor`, but also `--default-timeout` to a sensible value.
@disruptek : I tried a similar workaround using `--margin` definition but without success because it uses pixels so it can't play well with multihead setup and `--output=""`.
@emersion Allowing usage of value in % of the screen resolution could solve this issue though !

vilhalmer:
I think the implication wasn't that it was going to fundamentally change what last means, just that it will make it much harder to visually (or even logically) determine which notification is last. This would make it hard to control notifications with `makoctl`, for example.

TLogiviere:
Oh ok... So defining pseudo zones by means of redefining `margin` in criteria sections would be an easier solution... But it breaks with dual+ screens setups with different resolutions :scream: ! One solution would be to enable `margin` to use values in % the (focused or defined) `output` resolution.

Following this idea, `max-visible` would then be more useful if redefinable inside a criteria section, to reduce the risk of competition for display slots between defined pseudo zones. But it would be a bonus as I guess it would hit the same caveats as with `anchor` ?

joefiorini:
[...]

PR #228, opened 2020-02-01
Add capabilities to use multiple surfaces

Ongy:
This is very work in process and uses a rather nonsensical approach to
add surfaces into the system and configure them.
Outside this hardcoding, it should be somewhat fine for usage.

This addresses some of the comments in #123 and the current example is exactly about anchors.

### DIFF
--- a/config.c
+++ b/config.c
@@ -13,7 +13,6 @@
 #include "criteria.h"
 #include "types.h"
 
-
 static int32_t max(int32_t a, int32_t b) {
 	return (a > b) ? a : b;
 }
@@ -51,10 +50,6 @@ void init_default_config(struct mako_config *config) {
 	config->hidden_style.format = strdup("(%h more)");
 	config->hidden_style.spec.format = true;
 
-	config->output = strdup("");
-	config->layer = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
-
-	config->max_visible = 5;
 	config->max_history = 5;
 	config->sort_criteria = MAKO_SORT_CRITERIA_TIME;
 	config->sort_asc = 0;
@@ -63,9 +58,6 @@ void init_default_config(struct mako_config *config) {
 	config->button_bindings.right = MAKO_BINDING_DISMISS;
 	config->button_bindings.middle = MAKO_BINDING_NONE;
 	config->touch = MAKO_BINDING_DISMISS;
-
-	config->anchor =
-		ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
 }
 
 void finish_config(struct mako_config *config) {
@@ -76,7 +68,6 @@ void finish_config(struct mako_config *config) {
 
 	finish_style(&config->superstyle);
 	finish_style(&config->hidden_style);
-	free(config->output);
 }
 
 void init_default_style(struct mako_style *style) {
@@ -123,6 +114,13 @@ void init_default_style(struct mako_style *style) {
 	style->history = true;
 	style->icon_location = MAKO_ICON_LOCATION_LEFT;
 
+	style->output = strdup("");
+	style->layer = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
+	style->max_visible = 5;
+
+	style->anchor =
+		ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
+
 	// Everything in the default config is explicitly specified.
 	memset(&style->spec, true, sizeof(struct mako_style_spec));
 }
@@ -135,6 +133,7 @@ void finish_style(struct mako_style *style) {
 	free(style->icon_path);
 	free(style->font);
 	free(style->format);
+	free(style->output);
 }
 
 // Update `target` with the values specified in `style`. If a failure occurs,
@@ -145,6 +144,7 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	char *new_font = NULL;
 	char *new_format = NULL;
 	char *new_icon_path = NULL;
+	char *new_output = NULL;
 
 	if (style->spec.font) {
 		new_font = strdup(style->font);
@@ -168,6 +168,17 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 		if (new_icon_path == NULL) {
 			free(new_format);
 			free(new_font);
+			fprintf(stderr, "allocation failed\n");
+			return false;
+		}
+	}
+
+	if (style->spec.output) {
+		new_output = strdup(style->output);
+		if (new_output == NULL) {
+			free(new_format);
+			free(new_font);
+			free(new_icon_path);
 			fprintf(stderr, "allocation failed\n");
 			return false;
 		}
@@ -293,6 +304,27 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 		target->spec.border_radius = true;
 	}
 
+	if (style->spec.output) {
+		free(target->output);
+		target->output = new_output;
+		target->spec.output = true;
+	}
+
+	if (style->spec.anchor) {
+		target->anchor = style->anchor;
+		target->spec.anchor = true;
+	}
+
+	if (style->spec.layer) {
+		target->layer = style->layer;
+		target->spec.layer = true;
+	}
+
+	if (style->spec.max_visible) {
+		target->max_visible = style->max_visible;
+		target->spec.max_visible = true;
+	}
+
 	return true;
 }
 
@@ -385,28 +417,7 @@ bool apply_superset_style(
 
 static bool apply_config_option(struct mako_config *config, const char *name,
 		const char *value) {
-	if (strcmp(name, "max-visible") == 0) {
-		return parse_int(value, &config->max_visible);
-	} else if (strcmp(name, "max-history") == 0) {
-		return parse_int(value, &config->max_history);
-	} else if (strcmp(name, "output") == 0) {
-		free(config->output);
-		config->output = strdup(value);
-		return true;
-	} else if (strcmp(name, "layer") == 0) {
-		if (strcmp(value, "background") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND;
-		} else if (strcmp(value, "bottom") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM;
-		} else if (strcmp(value, "top") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
-		} else if (strcmp(value, "overlay") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY;
-		} else {
-			return false;
-		}
-		return true;
-	} else if (strcmp(name, "sort") == 0) {
+	if (strcmp(name, "sort") == 0) {
 		if (strcmp(value, "+priority") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_URGENCY;
 			config->sort_asc |= MAKO_SORT_CRITERIA_URGENCY;
@@ -419,29 +430,6 @@ static bool apply_config_option(struct mako_config *config, const char *name,
 		} else if (strcmp(value, "-time") == 0) {
 			config->sort_criteria |= MAKO_SORT_CRITERIA_TIME;
 			config->sort_asc &= ~MAKO_SORT_CRITERIA_TIME;
-		} else {
-			return false;
-		}
-		return true;
-	} else if (strcmp(name, "anchor") == 0) {
-		if (strcmp(value, "top-right") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP |
-				ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
-		} else if (strcmp(value, "top-center") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP;
-		} else if (strcmp(value, "top-left") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP |
-				ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT;
-		} else if (strcmp(value, "bottom-right") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM |
-				ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
-		} else if (strcmp(value, "bottom-center") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM;
-		} else if (strcmp(value, "bottom-left") == 0) {
-			config->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM |
-				ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT;
-		} else if (strcmp(value, "center") == 0) {
-			config->anchor = 0;
 		} else {
 			return false;
 		}
@@ -477,6 +465,8 @@ static bool apply_config_option(struct mako_config *config, const char *name,
 			return false;
 		}
 		return true;
+	} else if (strcmp(name, "max-history") == 0) {
+		return parse_int(value, &config->max_history);
 	}
 
 	return false;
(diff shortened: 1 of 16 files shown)


vilhalmer:
Thanks for working on this! I will hopefully have time to take an initial peek later in the week.

One thing to think about: how should this interact with pinning to specific outputs? Should you be able to specify that some notifications appear at `top-left` on `DP-1` and some at `top-left` on `DP-2` for example? And how does this affect the positioning of notifications that appear without a specific output, since there may be a surface visible on a non-focused output?

Ongy:
I mostly ignored groups for now and assumed they will only make sense when the entire group is on one surface and that doesn't change over the lifetime of affected notifiactions. Mostly to not have to understand them far enough.

For the second part, I'd think that surfaces should individually behave like the current one does.
So if the notification is using an existing surface, it will end up in that surface regardless of where the focus currently is.
If the surface is currently hidden or didn't get instanciated before, it will behave like it did so far.
Juding from your question, that's going to the focused output (if unconfigured and the output manager protocol is available) or to the configured output of that surface.

Ongy:
I'm not sure why I thought groups have anything to do with it. Sorry for that.

And yes, this would allow to have `top-left` `DP-1` and `top-left` `DP-2`. Or `top-right` `DP-2` if a user wants to.
Allocating notifications to the surfaces might be a bit tricky though.

My main motiviation is system notifications to be centered properly without some weird margins hacks, while others stay out of the way in the top right.

Ongy:
Ok, surfaces can be configured with the latest commit. Whether the mechanism how to do it is sane... I think it fits the current stuff at least somewhat^^

The harder question is how to assign a notification to a surfaces. And if it's done via styles, like I currentl have, how to handle merging them

Ongy:
Ok, I think this is mergible by now.

The one thing I'm unsure about is configuration, but I want input on that, so I won't consider it WIP anymore for now ;)

Oher thing, I think I'm leaking a buffer_pool on config reload (memcheck output), but I'm not sure how. If anyone knows, please tell me^^

config.c
```diff
@@ -51,21 +70,25 @@ void init_default_config(struct mako_config *config) {
...
 
 	config->button_bindings.left = MAKO_BINDING_INVOKE_DEFAULT_ACTION;
 	config->button_bindings.right = MAKO_BINDING_DISMISS;
 	config->button_bindings.middle = MAKO_BINDING_NONE;
 	config->touch = MAKO_BINDING_DISMISS;
+}
 
-	config->anchor =
-		ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
+static void free_surface_config(struct mako_surface_config *surface) {
```
vilhalmer, line 86: Style: this should use the `destroy_` prefix in line with other functions which free the given struct.

config.c
```diff
@@ -13,14 +13,33 @@
...
 
 static int32_t max(int32_t a, int32_t b) {
 	return (a > b) ? a : b;
 }
 
 void init_default_config(struct mako_config *config) {
 	wl_list_init(&config->criteria);
+	wl_list_init(&config->surfaces);
 	struct mako_criteria *new_criteria = create_criteria(config);
+	struct mako_surface_config *new_surface = create_surface(config);
```
vilhalmer, line 42: `create_surface` returns NULL if allocation fails but this isn't being handled here. We already do a mixed job of dealing with allocation failures in reasonable ways, so I'm not sure how careful we want to be here. I think it would be reasonable to bail all the way out if we fail this early.

main.c
```diff
@@ -79,11 +80,26 @@ static void finish(struct mako_state *state) {
...
 	finish_wayland(state);
 	finish_dbus(state);
 }
 
+static void init_surfaces(struct mako_state *state)
+{
+	wl_list_init(&state->surfaces);
+	struct mako_surface_config *config;
+	wl_list_for_each(config, &state->config.surfaces, link) {
+		/* We don't need the return value right now */
```
vilhalmer, line 98: Same question as above about how we should handle allocation failures here.

wayland.c
```diff
@@ -381,24 +387,25 @@ bool init_wayland(struct mako_state *state) {
...
-			strcmp(state->config.output, "") != 0) {
-		fprintf(stderr, "warning: configured an output but compositor doesn't "
-			"support xdg-output-unstable-v1 version 2\n");
+	if (state->xdg_output_manager == NULL) {
+		struct mako_surface *surface;
+		wl_list_for_each(surface, &state->surfaces, link) {
+			if (strcmp(surface->config->output, "") != 0) {
+				fprintf(stderr, "warning: configured an output "
+					"but compositor doesn't support "
+					"xdg-output-unstable-v1 version 2\n");
```
vilhalmer, line 396: I'd either break after logging this once or add the surface name to the message.

wayland.c
```diff
@@ -521,21 +533,22 @@ static void send_frame(struct mako_state *state) {
...
-		zwlr_layer_surface_v1_set_size(state->layer_surface,
+		zwlr_layer_surface_v1_set_size(surface->layer_surface,
 				style->width + style->margin.left + style->margin.right,
 				height);
-		zwlr_layer_surface_v1_set_anchor(state->layer_surface,
-				state->config.anchor);
-		wl_surface_commit(state->surface);
+		zwlr_layer_surface_v1_set_anchor(surface->layer_surface,
+				surface->config->anchor);
+				//state->config.anchor);
```
vilhalmer, line 550: Leftover.

config.c
```diff
@@ -636,6 +676,15 @@ int load_config_file(struct mako_config *config, char *config_arg) {
 			line[strlen(line) - 1] = '\0';
 		}
 
+		if (line[0] == '{' && line[strlen(line) - 1] == '}') {
```
vilhalmer, line 679: I'm going to have to ponder how I feel about adding a new type of section for this. I will admit no alternatives immediately spring to mind. Can you update the manpage with some examples of this so we can get a better idea of how it will feel?

config.c
```diff
@@ -13,14 +13,33 @@
 #include "criteria.h"
 #include "types.h"
 
+static struct mako_surface_config *create_surface(struct mako_config *config) {
```
vilhalmer, line 16: This should be called `create_surface_config` to match the struct, especially since a `mako_surface` is a different type.

surface.c
```diff
@@ -0,0 +1,38 @@
...
+
+struct mako_surface *create_surface(struct mako_state *state,
+		struct mako_surface_config *config) {
+	struct mako_surface *surface = calloc(1, sizeof(*surface));
+	if (!surface) {
+		return NULL;
+	}
+
+	surface->config = config;
+	surface->name = strdup(config->name);
```
vilhalmer, line 33: I don't think we should copy this since the config should always outlive any surface using it.

config.c
```diff
@@ -377,65 +404,73 @@ bool apply_superset_style(
...
-static bool apply_config_option(struct mako_config *config, const char *name,
-		const char *value) {
+static bool apply_surface_option(struct mako_surface_config *surface,
+		const char *name, const char *value) {
 	if (strcmp(name, "max-visible") == 0) {
-		return parse_int(value, &config->max_visible);
+		return parse_int(value, &surface->max_visible);
 	} else if (strcmp(name, "max-history") == 0) {
-		return parse_int(value, &config->max_history);
+		return parse_int(value, &surface->max_visible);
```
vilhalmer, line 412: Copypasta accident.

vilhalmer: More generally, do we want history to be per-surface? I don't think this code currently works as expected because we're still tracking history globally. I think it makes more sense to keep it global and provide mechanisms for excluding specific notifications (volume indicators, etc) from history entirely.

surface.c
Ongy: I thought so as well, until I got to reapply_config part.
That invalidates all configs, but not the surfaces currently.

I could tear down all surfaces and re-create them again, but I felt that it made more sense to keep them, and just cleanup. Both from ease of adding it to codeflow, and a (micro)optimization.

config.c
Ongy: TBH. I mostly rebased onto that syntactically.

Probably has the history on `mako_state` while the max_history is currently on the surface, because that's how git auto-merged.

I'll have to get a grasp of history in general. Would it make sense to have the exclusion on the surface, or per notification?

Ongy: Ahh, I just saw that there's a style option for it.

config.c
Ongy: I'd rather not right now. The line directly above it (`create_criteria`) has the exact same issue.

Fixing that in a sane way requires changing the return type of the function. While I don't mind doing it, I'd rather outsource it into a different PR based on this, which also incorporates valgrind memcheck and clang scan-build based changes.

main.c
Ongy: In practice I'd probably abort the daemon if it's the default `"(root)"` surface, and otherwise continue on, since it'll still work, just not quite as expected.

Ongy:
Thanks for the feedback. I incorporated the changes that are easy right now.

I'll get to the examples soon. For the others I think we need a bit of clarification.

Ongy:
Ok, I added the stuff to the man page.

Also fixed a weirdness while thinking about it again.
The surface configuration as criteria option made no sense, so I moved it to style.

I added my configuration as example config as well. Not 100% sure if that's what you intended.

surface.c
vilhalmer: Hmm, I see. We do our best to immediately apply changes to styles, I wonder if this should include moving notifications? If so then we'd need to tear them all down in case things moved.

vilhalmer:
>I added my configuration as example config as well. Not 100% sure if that's what you intended.

This makes it much easier to visualize, thanks!

surface.c
Ongy: Hm, I think we are thinking about slightly different things.

So the current code order is:
1. Deallocate configs
2. Re-read configuration
3. Ensure every configured surface exists
4. Reapply styles, which re-evaluates the surface
5. Trigger redraw / de-allocate unused surfaces

The new step is 3, and the added parts to 4 and 5 of course.
Everything still happens at the same time. They don't change the timing (outside of course pure computation time) at all.

main.c
Ongy: See https://github.com/emersion/mako/pull/264

vilhalmer:
I've been thinking about whether it makes sense to add a new type of configuration section for this and I'm leaning toward no. Introducing named surfaces also introduces a need to perform an extra level of validation: we don't really want to allow the user to create two surfaces with the same position, as the surfaces would fight for that location. Worse, the exact behavior of this would depend on the compositor's implementation of layer-shell. (Not to mention the possiblity of referencing a layer that wasn't configured.)

I think instead it would make sense to keep the layer options as style options, and have mako figure out which surfaces to create and where to place things based on the set of `(output, layer, anchor)`. While this is slightly more verbose in the config file, there are only a couple options and I think it would be just as intuitive for the user. How do you feel about this?

The downside to this is that I'm not sure what behavior it implies for `max-visible`. @emersion do you have any opinions on this?

Ongy:
Validating the config doesn't specify non-existant surfaces and error (and warn for unused surfaces) would probably be a good idea.

I went with this approach over dynamically generating the surfaces based on (what I called) `SURFACE OPTIONS` in styles, because that requires generating surfaces on the fly when notifications come in, which I'm not a big fan of.
There's also an issue about allowing notifications on multiple screens. A set of surfaces is at least possible with this approach (though not exactly trivial), and becomes harder with auto-generation/detection.


And with weird margins the user can still configure notifications that overlap that wouldn't be detected properly, e.g. Right top + right Bottom could overlap if enough notifications come in.
Unless we want to spend the work detecting that, then wrapping them into a single surface and maybe untangling after some got closed. Which I don't feel like doing^^

vilhalmer:
>that requires generating surfaces on the fly when notifications come

This isn't necessarily true, mako would still know all of the surfaces which could possibly exist at the time of reading the config.

Ongy:
Hm, I might look into it tomorrow and write up a PoC for comparison.
I know that we can in theory compute a superset of possible surfaces, I was just thinking that might be too much of an over-estimate.

But thinking about it more and looking at the 3 variables that go into it... Probably not an issue.

Ongy:
Ok, so the latest commit tries to apply the logic you suggested.
It's not exactly pretty IMO from a code/logic perspective, but might make more sense from the config side.

And max_visible is somewhat arbitrary right now, as it's taked from the fully computed style that first creates the surface config.

config.c
```diff
@@ -377,65 +446,74 @@ bool apply_superset_style(
 	return true;
 }
 
-static bool apply_config_option(struct mako_config *config, const char *name,
-		const char *value) {
+static bool apply_surface_option(struct mako_style *surface,
```
vilhalmer, line 449: Looks like this can be merged back into `apply_style_option` now since they get moved over to the surface configs later.

Ongy:
Like said on IRC, I'll try the suggestion to generate them on the fly, probably this weekend.

Then we have 3 approaches in this PR as code in commits, and once we decide on the way to go, I'll do some code cleanup :)
And then probably clean up the cleanup PR o.0

vilhalmer:
Thanks for your work on this! I'm excited to put it to use.

Ongy:
Ok, honestly I'm unsure why I didn't go for this from the get go, looking at it now.
Probably had some weird ideas about corner cases....

Outside the changes needed to have multi-surface support at all, this is quite simple on the code side and makes things easy to use as well.


This is still unsanitized from older stuff. There's some functions that need removing (create_surface_config in favor of configured), and I don't think we really need `struct mako_surface_config` at all. It's only used as aggregate argument to `create_surface` now either way.

Future work:
* Figure out the monitor we want to show the notification on before surface resolution is done, to merge configured monitor/defaulted montiro surfaces.
* Cleanup
* Remove name for direct comparisons (it's was just an easy aggregate to compare 3 values)

config.c
```diff
@@ -548,7 +591,10 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 			style->padding.right = max(style->border_radius, style->padding.right);
 		}
 		return spec->border_radius;
-	}
+	} /*else if (strcmp(name, "surface") == 0) {
+		free(style->surface);
+		return spec->surface = !!(style->surface = strdup(value));
+	} */
```
vilhalmer, line 597: Leftover comment.

dbus/mako.c
```diff
@@ -230,12 +236,31 @@ static int handle_list_notifications(sd_bus_message *msg, void *data,
...
+ * The way surfaces are re-build here is not quite intuitive.
+ * 1. We set the surface->config pointers to NULL
+ *    The previously pointed to values got free()ed before!
+ * 2. We iterate the new surface configs. Set the existing surfaces' config
+ *    pointers to our current configuration and create new surfaces when necessary.
+ * 3. We apply configurations to existing notifications.
+ * 4. We iterate surfaces again. Here we check for surfaces that don't have a
+ *    config set (surface->config == NULL), those can be discarded.
+ *    Configs that have a config (i.e. are still in use) are scheduled for redraw.
+ */
```
vilhalmer, line 249: This comment is out of date.

mako.5.scd
```diff
@@ -289,6 +290,29 @@ associated command-line option.
...
+	Default: ""
+
+# SURFACES
+
+In addition to the set of options at the top of the file, the config file may
+contain zero or more sections, each containing any combination of the
+*SURFACE OPTIONS*. The sections, called surfaces, are defined with a squiggly
+square bracket syntax. The brackets contain the name of the surface:
+
+	\{internal\}
```
vilhalmer, line 309: This bit can be removed again.

mako.5.scd
```diff
@@ -344,6 +368,22 @@ Maintained by Simon Ser <contact@emersion.fr>, who is assisted by other
...
+++
+\[category=system\]++
+surface=system++
+background-color=#28557780++
+margin=0,0,250++
+width=120++
+text-color=#efefef++
+border-radius=20
+
+
```
vilhalmer, line 386: Ditto.

config.c
```diff
@@ -740,6 +777,7 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 		{"max-visible", required_argument, 0, 0},
 		{"max-history", required_argument, 0, 0},
 		{"history", required_argument, 0, 0},
+		{"surface", required_argument, 0, 0},
```
emersion, line 780: This is a leftover from previous code since the `surface` option has been removed, correct?

mako.5.scd
```diff
@@ -298,6 +297,15 @@ associated command-line option.
 
 	Default: 0
 
+*surface* = _surface_
```
emersion, line 300: This is a leftover from previous code since the `surface` option has been removed, correct?

mako.5.scd
```diff
@@ -214,12 +184,41 @@ _none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_,
 
 *group-by* = _field[,field,...]_
 	A comma-separated list of criteria fields that will be compared to other
-	visible notifications to determine if this one should form a group with
-	them. All listed criteria must be exactly equal for two notifications to
+	visible notifications to determine if this one should form a group with them. All listed criteria must be exactly equal for two notifications to
```
emersion, line 187: Style: wrap to 80 chars

config.c
```diff
@@ -385,28 +417,8 @@ bool apply_superset_style(
...
-		} else if (strcmp(value, "top") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
-		} else if (strcmp(value, "overlay") == 0) {
-			config->layer = ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY;
-		} else {
-			return false;
-		}
-		return true;
-	} else if (strcmp(name, "sort") == 0) {
+
```
emersion, line 420: Style nit: stray newline

render.c
```diff
@@ -319,9 +319,10 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	return notif_height;
 }
 
-int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
+int render(struct mako_state *state, struct mako_surface *surface, struct pool_buffer *buffer, int scale) {
 	struct mako_config *config = &state->config;
 	cairo_t *cairo = buffer->cairo;
+	size_t count = 0;
```
emersion, line 325: Style: move this declaration right before the `wl_list_for_each`

include/render.h
```diff
@@ -2,7 +2,8 @@
 #define MAKO_RENDER_H
 
 struct mako_state;
+struct mako_surface;
 
-int render(struct mako_state *state, struct pool_buffer *buffer, int scale);
+int render(struct mako_state *state, struct mako_surface *surface, struct pool_buffer *buffer, int scale);
```
emersion, line 7: Since the state is already stored in the `mako_surface` struct, no need to have it as a separate argument anymore

PR #229, opened 2020-02-06
Timeout = 0, max-visible = 0

arkdchst:
Now works with zero timeout as it should be, it's more intuitively.

Before it with _max-visible=0_ one notification was showing however. It was equivalent to _max-visible=1._
Now it works with _max-visible=0,_ so all notifications can be hidden.


arkdchst:
image
Sorry, I didn't see this.

PR #230, opened 2020-02-06
max-visible = 0

arkdchst:
Before it with _max-visible=0_ one notification was showing however. It was equivalent to _max-visible=1._
Now it works with _max-visible=0,_ so all notifications can be hidden.


render.c
```diff
@@ -275,6 +275,12 @@ int render(struct mako_state *state, struct pool_buffer *buffer, int scale) {
 	int pending_bottom_margin = 0;
 	struct mako_notification *notif;
 	wl_list_for_each(notif, &state->notifications, link) {
+
```
emersion: Style: unnecessary newline

emersion:
Can you squash these commits together?

arkdchst:
Done.

Issue #213, opened 2020-01-05
Cannot deindent in literal block


varac:
Current master from toady fails to build for me:

```
❯ meson setup --wipe build
The Meson build system
Version: 0.51.2
Source dir: /home/varac/projects/wayland/mako
Build dir: /home/varac/projects/wayland/mako/build
Build type: native build
Project name: mako
Project version: 1.3.0
C compiler for the host machine: cc (gcc 9.2.1 "cc (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Run-time dependency cairo found: YES 1.16.0
Run-time dependency pango found: YES 1.42.3
Run-time dependency pangocairo found: YES 1.42.3
Run-time dependency glib-2.0 found: YES 2.62.1
Run-time dependency gobject-2.0 found: YES 2.62.1
Library rt found: YES
Run-time dependency wayland-client found: YES 1.17.0
Run-time dependency wayland-protocols found: YES 1.18
Run-time dependency libsystemd found: YES 242
Run-time dependency gdk-pixbuf-2.0 found: YES 2.40.0
Program wayland-scanner found: YES (/usr/bin/wayland-scanner)
Configuring fr.emersion.mako.service using configuration
Program scdoc found: YES (/usr/bin/scdoc)
Program sh found: YES (/bin/sh)
Build targets in project: 4
Found ninja-1.9.0 at /usr/bin/ninja

❯ ninja -C build          
ninja: Entering directory `build'
[11/31] Generating makoctl.1 with a custom command.
FAILED: makoctl.1 
/bin/sh -c '/usr/bin/scdoc < ../makoctl.1.scd > makoctl.1'
Error at 40:4: Cannot deindent in literal block
[16/31] Compiling C object 'mako@exe/event-loop.c.o'.
ninja: build stopped: subcommand failed.

```

Ubuntu 19.10, scdoc 1.9.6-1

varac:
Turns out that upgrading scdoc to 1.10.0-1 from ubuntu 20.04 fixes this.

Please add this dependency to the README.md.

emersion:
Please send a patch.

emersion:
[1.9.7](https://git.sr.ht/~sircmpwn/scdoc/refs/1.9.7) should fix it too. Should get in the Ubuntu repos, since it's a bugfix release.

PR #232, opened 2020-02-10
makoctl.1.scd: deindent literal block

0xdc:
commit e6a8cadab added an indented literal block that scdoc couldn't
deindent.

  $ scdoc <makoctl.1.scd
  Error at 40:4: Cannot deindent in literal block

Deindenting the block fixes the issue.

### DIFF
--- a/makoctl.1.scd
+++ b/makoctl.1.scd
@@ -36,10 +36,10 @@ Sends IPC commands to the running mako daemon via dbus.
 
 	Examples:
 
-		```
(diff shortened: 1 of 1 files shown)


emersion:
What is you scdoc version?

0xdc:
app-text/scdoc-1.9.3-r1 from gentoo

emersion:
You need to upgrade mako. See https://github.com/emersion/mako/issues/213.

PR #233, opened 2020-02-10
. is included in link when followed

morten-b:


emersion:
Hmm. I'd say this is a terminal emulator bug.

morten-b:
As far as I can tell a trailing dot is a valid adress, so the terminal emulator have no way of knowing if it is part of the link or the language syntax.

See this: http://www.dns-sd.org/TrailingDotsInDomainNames.html

emersion:
I guess the real solution would be to add support for this in `scdoc`. Opened https://todo.sr.ht/~sircmpwn/scdoc/25

morten-b:
Excellent. Should we close this?

PR #237, opened 2020-03-03
Completions: add fish-shell

ammgws:


ammgws:
Ah I forgot about makoctl

Issue #234, opened 2020-02-12
Feature request: allow dismissing grouped messages in one click


grahamc:
I have some programs which send me many messages, more than I'd like and in a way I can't easily fix.

Sometimes my notifications look something like this (though this is exaggerated):
 
![image](https://user-images.githubusercontent.com/76716/74368481-bf227000-4da1-11ea-9a99-b9d0db8883df.png)

I'd like to be able to say yes, group by `app-name,summary` and also when I click to ack the notification, dismiss all of them.  Right now I either clear it with the CLI, or by clicking the notification until they are all gone.

vilhalmer:
I feel your pain, I have this same problem with wee-slack. (Though it makes for a good performance test, I suppose.)

emersion:
We could probably add `MAKO_BINDING_DISMISS_GROUP` here:

https://github.com/emersion/mako/blob/175b8d8ab752bcc29da39ff89eca0b274af66628/notification.c#L300

qguv:
Was this fixed by #224? If so, this can be closed :slightly_smiling_face:

emersion:
Indeed, fixed by #244!

PR #244, opened 2020-03-20
Support dismissing notifications by group

grahamc:
Implements #234.

For all practical purposes, I don't write C, so... my apologies if this is bad.

Locally I tested with this:

```diff
diff --git a/config.c b/config.c
index b337895..1f6e467 100644
--- a/config.c
+++ b/config.c
@@ -60,7 +60,7 @@ void init_default_config(struct mako_config *config) {
        config->sort_asc = 0;

        config->button_bindings.left = MAKO_BINDING_INVOKE_DEFAULT_ACTION;
-       config->button_bindings.right = MAKO_BINDING_DISMISS;
+       config->button_bindings.right = MAKO_BINDING_DISMISS_GROUP;
        config->button_bindings.middle = MAKO_BINDING_NONE;
        config->touch = MAKO_BINDING_DISMISS;
```

I looked at making this properly configurable, but that seemed like a reach. If this PR looks good, perhaps I'll take a look at that.


notification.c
```diff
@@ -135,6 +135,28 @@ struct mako_notification *get_notification(struct mako_state *state,
...
 
+void close_group_notifications(struct mako_notification *top_notif,
+	       enum mako_notification_close_reason reason) {
+	struct mako_notification *notif, *tmp;
+
+	struct mako_criteria *notif_criteria = create_criteria_from_notification(
+			top_notif, &top_notif->style.group_criteria_spec);
+
+	wl_list_for_each_safe(notif, tmp, &top_notif->state->notifications, link) {
+		if (notif->id == top_notif->id) {
```
emersion: The top notification may be ungroupped. In which case we probably want to just close it as you've done.

Why do we need this special case though? Can't we just treat the first notification just like other notifications and close it in this loop?

notification.c
```diff
@@ -305,6 +327,9 @@ void notification_execute_binding(struct mako_notification *notif,
 	case MAKO_BINDING_DISMISS:
 		close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
 		break;
+	case MAKO_BINDING_DISMISS_GROUP:
+		close_group_notifications(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
```
emersion: Maybe this function should take `notif->state` instead of `notif`.

notification.c
```diff
@@ -135,6 +135,28 @@ struct mako_notification *get_notification(struct mako_state *state,
...
+			continue;
+		}
+		if (!match_criteria(notif_criteria, notif)) {
+			continue;
+		}
+
+		close_notification(notif, reason);
+	}
+
+	free(notif_criteria);
```
emersion: We should use `destroy_criteria` here instead.

notification.c
grahamc: That seemed to cause a segfault in match_criteria when strcmp'ing against app names, I think. Not sure exactly, but it seemed create_criteria_from_notification assumes the notif isn't being free'd before the criteria is free'd.

notification.c
grahamc: I think I need notif in order to create the criteria?

notification.c
vilhalmer: Hmm. Nothing inside the criteria struct should be pointing at the notification, all the strings are copied.

notification.c
vilhalmer: Yikes, the existing close_notification uses free directly as well. I bet that's at least one of our outstanding memory leaks. My bad. -_-'

emersion:
>I do have one more architecture-level thought: every call to close_notification triggers a regroup, which is pointless when we're about to close all of them. It would be nice to avoid that. I wonder if these two functions could be combined and have a close_group flag instead, so that it can skip the extra regroups?

If it doesn't make the code too complex. This part is already hard to grok.

notification.c
emersion: Ah, indeed!

notification.c
grahamc: Thanks, fixed up this and the other case.

notification.c
grahamc: You're right, it seems I see the segfault for unrelated reasons.

grahamc:
So I tried fixing this, but I ran out of ideas. Along the way I changed the `free`s to `destroy_criteria`'s and that caused more segfaults -- which I think are solved by https://github.com/emersion/mako/pull/247. I've rebased this PR on to that PR. I wonder if the segfault has to do with regrouping and wl_list_for_each_safe's rules about how you can modify the list being iterated on. That said, I'm sort of stumbling around in the dark here.

grahamc:
The segfault I saw was in `match_criteria`, 

```
	if (spec.app_name &&
			strcmp(criteria->app_name, notif->app_name) != 0) {
		return false;
	}
```

with notif->app_name being inaccessible memory.

grahamc:
I applied the patch you sent on IRC, and this seems to work fine now.

emersion:
All right, this LGTM. Can you squash all of these commits into a single one?

PR #245, opened 2020-03-22
dbus: Request name async

kennylevinsen:
If a method call towards mako caused service activation, the activating call would not be answered and simply time out. Switching the name request to be async allows the activating call to be serviced.

Note that notify-send would not display this issue, as it explicitly calls org.freedesktop.DBus.StartServiceByName on startup. A direct org.freedesktop.Notifications.Notify call is needed, such as the one from https://git.sr.ht/~kennylevinsen/poweralertd.

### DIFF
--- a/dbus/dbus.c
+++ b/dbus/dbus.c
@@ -29,7 +29,7 @@ bool init_dbus(struct mako_state *state) {
 		goto error;
 	}
 
-	ret = sd_bus_request_name(state->bus, service_name, 0);
+	ret = sd_bus_request_name_async(state->bus, NULL, service_name, 0, NULL, NULL);
 	if (ret < 0) {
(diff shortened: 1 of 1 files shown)


emersion:
I wonder why `_async` would make D-Bus activation work. I'm worried to rely on a race condition here.

kennylevinsen:
Okay, I have found out why sync didn't work: It uses sd_bus_call internally, which reads messages into a queue without processing them until it finds what it was looking for. This means that sd_bus_process needs to be called to process this queue despite there being nothing on the socket to wake up our poll. This part is easily fixable.

However, there are additional issues that were automatically fixed by the async implementation, as the notification currently still fails to show. Still trying to figure out why this might be different between the two.

PR #248, opened 2020-03-23
config: allow configuring button and touch bindings

grahamc:


mako.5.scd
```diff
@@ -59,6 +59,25 @@ Empty lines and lines that begin with # are ignored.
 
 	Default: top-right
 
+# BUTTON AND TOUCH BINDINGS
+
+Specify the action to take per button or touch event. Supported values are
+_none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_,
+and _bottom-center_.
```
emersion: Is "bottom-center" a leftover from a copy-pasta?

emersion:
Otherwise looks good!

mako.5.scd
grahamc: d'oh, yep, I even have an uncommitted, staged patch here fixing that.

grahamc:
Fixup'd!

PR #255, opened 2020-04-06
Add a titlebar

mendelmaleh:
I wanted mako to match the sway window aesthetics, so I added a titlebar. To enable it `title-format` needs to be set in the config.
It will only render if that option is set and there is text to show, otherwise the notification will look like it used to before this.


mendelmaleh:
Force pushed because I fixed a bug in the first commit and rebased.

emersion:
Thanks for the patch! I'm wondering if this feature is worth the added complexity...

mendelmaleh:
Fixed up some quirks with border radius. I'm not aware of any other bugs, after using this for around a week and testing a bit. I do think this is worth it, it fits really well with sway, imo ¯\_(ツ)_/¯

emersion:
Would there be a nice way to de-duplicate some of the rendering logic? `format` and `title_format` are pretty similar.

mendelmaleh:
I will try, but I'm very much a beginner at C. For the most part I avoided touching what I didn't need to xD

mendelmaleh:
Another edge case done, I will try now deduplicating the config.c.

mendelmaleh:
@emersion does this last commit look any good? It compiles and runs fine, but I really don't know much about styling, naming, etc.
I didn't see a code style for the repo, so if there is a more general one I should be aware of do tell me, all resources are appreciated.

ammgws:
@mendelmaleh Perhaps this will help regarding the styling: https://github.com/swaywm/sway/blob/master/CONTRIBUTING.md

emersion:
I'm on the fence wrt. this PR: I'm not sure we should add this extra functionality. I don't want to maintain software with too many features, and this opens the door to other UI elements like footer/sidebars/etc.

vilhalmer:
I wonder, would it be sufficient to instead allow specifying a gradient with multiple stops that get passed straight to cairo? That would avoid adding (much) extra work on the mako side, and you can create a hard edge that appears to be a title bar if you wish.

markstos:
I think this is unnecessary. Now the PR has been open almost two years and there's been little interest in it.

NiciTheNici:
I feel like this merge request can be closed after 2 years of no activity as well. @emersion

noahhefner:
Probably safe to close this now?

Issue #259, opened 2020-04-18
Mako service fails to load due to ExecCondition


grumpey:
Fedora 32
bash 5.0.11
mako 1.4.1 up to e94407e 

Remarking out ExecCondition=/bin/sh -c '[ -n "$WAYLAND_DISPLAY" ]' or changing it to ExecCondition=/bin/sh -c '[ -n $WAYLAND_DISPLAY ]  in the service file appears to resolve. 

Thanks

PR #260, opened 2020-04-19
Remove double quotes around variable in service file

grumpey:
Fixes https://github.com/emersion/mako/issues/259 on Fedora, would need to be verified on other distros.


emersion:
Why does it fix the issue? Double-quotes should work.

grumpey:
Closing, worked on another distro. Thanks

PR #261, opened 2020-04-19
Disable antialiasing when requested by the output

### LINKED ISSUE #258: WL_OUTPUT_SUBPIXEL_NONE should disable subpixel
See [sway/#5228](https://github.com/swaywm/sway/issues/5228#issuecomment-615838382): Mako currently treats the `WL_OUTPUT_SUBPIXEL_NONE` setting with the default subpixel orientation. As such, there is no way to disable subpixel on Mako (i.e. set it to grayscale).

[Link](https://github.com/emersion/mako/blob/175b8d8ab752bcc29da39ff89eca0b274af66628/render.c#L59) to relevant code (thanks to progandy).

I personally strongly dislike the colored fringes created by all types of subpixel rendering and would really appreciate a way to switch it to grayscale/gradient antialiasing. Note that I have also set grayscale in fontconfig, which Mako doesn't respect:

```
$ file /etc/fonts/conf.d/10-no-sub-pixel.conf
/etc/fonts/conf.d/10-no-sub-pixel.conf: symbolic link to ../conf.avail/10-no-sub-pixel.conf
```

vilhalmer:
~Not tested yet, but this seems like the correct way to disable subpixel in cairo.~

Fixes #258.

### DIFF
--- a/render.c
+++ b/render.c
@@ -80,9 +80,13 @@ static void set_font_options(cairo_t *cairo, struct mako_state *state) {
 	}
 
 	cairo_font_options_t *fo = cairo_font_options_create();
-	cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_SUBPIXEL);
-	cairo_font_options_set_subpixel_order(fo,
-		get_cairo_subpixel_order(state->surface_output->subpixel));
+	if (state->surface_output->subpixel == WL_OUTPUT_SUBPIXEL_NONE) {
+		cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_GRAY);
+	} else {
+		cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_SUBPIXEL);
+		cairo_font_options_set_subpixel_order(fo,
+				get_cairo_subpixel_order(state->surface_output->subpixel));
+	}
 	cairo_set_font_options(cairo, fo);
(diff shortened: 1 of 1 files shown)


vilhalmer:
I figured out how to test it. :)

render.c
```diff
@@ -80,9 +80,13 @@ static void set_font_options(cairo_t *cairo, struct mako_state *state) {
 	}
 
 	cairo_font_options_t *fo = cairo_font_options_create();
-	cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_SUBPIXEL);
-	cairo_font_options_set_subpixel_order(fo,
-		get_cairo_subpixel_order(state->surface_output->subpixel));
+	if (state->surface_output->subpixel == WL_OUTPUT_SUBPIXEL_NONE) {
+		cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_NONE);
```
emersion, line 84: This disables anti-aliasing. We shouldn't do this.

emersion:
`WL_OUTPUT_SUBPIXEL_NONE` doesn't mean "disable anti-aliasing.

vilhalmer:
Ah, I misunderstood.

vilhalmer:
Updated to use grayscale instead of disabling antialiasing entirely.

vilhalmer:
Ah, should change the commit message…

PR #266, opened 2020-05-01
Icon location

Ongy:
Adds the possibility to specify the icon location as style option.

Example screenshots:
bottom
left
right
top

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -20,13 +20,21 @@ enum mako_sort_criteria {
 	MAKO_SORT_CRITERIA_URGENCY = 2,
 };
 
+enum mako_icon_location {
+	MAKO_ICON_LOCATION_LEFT,
+	MAKO_ICON_LOCATION_RIGHT,
+	MAKO_ICON_LOCATION_TOP,
+	MAKO_ICON_LOCATION_BOTTOM,
+};
+
 // Represents which fields in the style were specified in this style. All
 // fields in the mako_style structure should have a counterpart here. Inline
 // structs are also mirrored.
 struct mako_style_spec {
 	bool width, height, margin, padding, border_size, border_radius, font,
 		markup, format, actions, default_timeout, ignore_timeout, icons,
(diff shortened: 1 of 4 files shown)


fourstepper:
Love it!

render.c
```diff
@@ -228,19 +259,46 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
...
-			(text_x - icon->width) / 2;
-		double ypos = offset_y + style->border_size +
-			(notif_height - icon->height - border_size) / 2;
+		double xpos = -1;
+		double ypos = -1;
+		switch (style->icon_location) {
+		case MAKO_ICON_LOCATION_LEFT:
+			xpos = offset_x + style->border_size +
+				(text_x - icon->width) / 2;
+			ypos = offset_y + style->border_size +
```
emersion, line 268: Can we make this a little bit easier to read by storing this value in a `ypos_center` variable before the switch? Same for `xpos_center`.

render.c
```diff
@@ -98,30 +98,50 @@ static int render_notification(cairo_t *cairo, struct mako_state *state,
 	int padding_height = style->padding.top + style->padding.bottom;
 	int padding_width = style->padding.left + style->padding.right;
 	int radius = style->border_radius;
+	// I like micro optimizatins
+	bool icon_vertical = style->icon_location & 0x02;
```
emersion, line 102: Yeah, let's not

include/config.h
```diff
@@ -20,13 +20,21 @@ enum mako_sort_criteria {
 	MAKO_SORT_CRITERIA_URGENCY = 2,
 };
 
+enum mako_icon_location {
+	MAKO_ICON_LOCATION_LEFT = 0,
```
emersion, line 24: Nit: no need to define the values explicitly.

Ongy:
The `ypos_center` and `xpos_center` could be assigned to the actual `xpos/ypos` as well, but I think it's slightly more semantic this way.

PR #267, opened 2020-05-01
Improve center alignment

Ongy:
Title says all. See attached screenshots for before and after.

Or, probably doesn't. Quick edit^^
So this happens if there's a center alignment and notifications with different widths. This should fix that. Probably more important with the multi-surface stuff, but also a general fix.
center-align
center-patched


vilhalmer:
Thanks! Sorry for the delay, this email got lost in a pile. :)

PR #268, opened 2020-05-04
Changes in order to expose various config options in the README

fourstepper:


emersion:
Sorry, I don't want to provide a default configuration file.

fourstepper:
Sounds fair. I have deleted the example file and the association in the README. I think exposing the manuals directly in the README is a good idea, though

emersion:
Thanks!

Issue #38, opened 2018-05-23
Support for running abitrary scripts on notification


adrubesh:
Taken from my previous issue on playing a notification chime. Perhaps the same rules for #3 could be used to run user-defined scripts so that they could be notification/urgency/etc specific?

emersion:
Could be something like:

```ini
[urgency=critical]
command=mpv ~/critical.ogg
```

Where `command` would be executed with `/bin/sh`.

vilhalmer:
There is an explicit "sound" capability in the spec which specifies hints that must be supported, perhaps we should support this as well and then offer some way to expand the content of hints in a format to pass as arguments to the exec'd tool?

emersion:
If we add the "sound" capability, it means we must support the `sound-file` hint (and `sound-name`). So we could have something like:

```ini
# %S (or something else) expands to sound-file
exec-sound=mpv %S

[urgency=critical]
default-sound=~/critical.ogg
```

vilhalmer:
I was picturing something more like:
```
[urgency=critical]
exec=mpv %H(sound-file)
```
to keep it from being special, but I can see it would be tricky to figure out what to do if the app passed `sound-name` vs `sound-file`.

emersion:
Indeed that's a much better idea. And we could just scan formats for `%H(sound-file)` to decide whether we should expose the "sound" capability or not.

IMHO we should always expand `sound-name` to get a full filepath, not sure how to do that though.

vilhalmer:
The [sound naming spec](http://0pointer.de/public/sound-naming-spec.html) doesn't say anything about resolving them to paths, on Arch at least they appear to live in /usr/share/sounds/freedesktop. I'm not sure if it's worth supporting sound-name or not, the notification spec only mentions `sound-file` and `suppress-sound` as required:
>The server supports sounds on notifications. If returned, the server must support the "sound-file" and "suppress-sound" hints.

emersion:
Yeah. We could just decide not to support it, check if any apps in the wild use it, and consider adding support if we find any.

vilhalmer:
Ah, I found it in a related spec: http://0pointer.de/public/sound-theme-spec.html#directory_layout

jeremejevs:
I'd like to mention a usecase: timers / alarms. It would be nice to be able to play a looping sound, until the notification expires or gets dismissed.

[...]

Issue #36, opened 2018-05-22
Playing a sound on notifications


adrubesh:
Are notification chimes something we'd like to implement? Maybe they could be configured based on application with #3?

SirCmpwn:
-1

SirCmpwn:
Maybe just run an arbitrary command when a notification comes in?

emersion:
dunst supports scripts. I've been wondering about supporting scripts too.

adrubesh:
Yeah, the script support would support a much wider range of use cases. I like that.

PR #270, opened 2020-05-06
Use synchronous hints for replacing notifications

williamvds:
`synchronous`, `private-synchronous`, `x-canonical-private-synchronous`, and
`x-dunst-stack-tag` are supported.
When a new notification with a tag is created the previous one with the
same tag is closed with `MAKO_NOTIFICATION_CLOSE_REPLACED`, and the new
notification is added to the list.

There's a behavior difference between replacing by ID and the new stack tag method. Replacing by ID will modify the notification in-place, whereas with the tag the new notification can be positioned differently as it's separately inserted into the list.
The synchronous tag is only identified once other hints have been processed and values have been set it isn't as easy to replace in-place.

Personally I prefer updated notifications being bumped up if they're replaced when they're supposed to be sorted by time, and I think replacing by ID should do the same. But if it's desirable to maintain the current behavior, here's some possible approaches:

1. Separate the notification ID generation from `create_notification`, so the ID can be set to the replaced ID or a new one as necessary
2. Find the tag hint before creating/replacing a notification. Not sure if the D-Bus API allows going through containers multiple times like this
3. Steal the existing notification's ID anyway (and optionally don't worry about the newly generated ID being lost)

### DIFF
--- a/dbus/xdg.c
+++ b/dbus/xdg.c
@@ -59,6 +59,16 @@ static int handle_get_capabilities(sd_bus_message *msg, void *data,
 		}
 	}
 
+	ret = sd_bus_message_append(reply, "s", "x-canonical-private-synchronous");
+	if (ret < 0) {
+		return ret;
+	}
+
+	ret = sd_bus_message_append(reply, "s", "x-dunst-stack-tag");
+	if (ret < 0) {
+		return ret;
+	}
+
 	ret = sd_bus_message_close_container(reply);
 	if (ret < 0) {
 		return ret;
@@ -239,6 +249,14 @@ static int handle_notify(sd_bus_message *msg, void *data,
 			// will win over app_icon if provided.
 			free(notif->app_icon);
 			notif->app_icon = strdup(image_path);
+		} else if (strcmp(hint, "x-canonical-private-synchronous") == 0 ||
+				strcmp(hint, "x-dunst-stack-tag") == 0) {
+			const char *tag = NULL;
+			ret = sd_bus_message_read(msg, "v", "s", &tag);
+			if (ret < 0) {
+				return ret;
+			}
+			notif->tag = strdup(tag);
 		} else if (strcmp(hint, "image-data") == 0 ||
 				strcmp(hint, "image_data") == 0 ||  // Deprecated.
 				strcmp(hint, "icon_data") == 0) {  // Even more deprecated.
@@ -342,6 +360,17 @@ static int handle_notify(sd_bus_message *msg, void *data,
 		return ret;
 	}
 	notif->requested_timeout = requested_timeout;
+
+	if (notif->tag) {
+		// Find and replace the existing notfication with a matching tag
+		struct mako_notification *replace_notif = get_tagged_notification(state, notif->tag, app_name);
+		if (replace_notif) {
+			notif->id = replace_notif->id;
+			wl_list_insert(&replace_notif->link, &notif->link);
+			destroy_notification(replace_notif);
+			replaces_id = notif->id;
+		}
+	}
 
 	// We can insert a notification prior to matching criteria, because sort is
 	// global. We also know that inserting a notification into the global list
--- a/include/notification.h
+++ b/include/notification.h
@@ -42,6 +42,7 @@ struct mako_notification {
 	enum mako_notification_urgency urgency;
 	char *category;
 	char *desktop_entry;
+	char *tag;
 	int32_t progress;
 	struct mako_image_data *image_data;
 
@@ -89,6 +90,7 @@ char *format_hidden_text(char variable, bool *markup, void *data);
 char *format_notif_text(char variable, bool *markup, void *data);
 size_t format_text(const char *format, char *buf, mako_format_func_t func, void *data);
 struct mako_notification *get_notification(struct mako_state *state, uint32_t id);
+struct mako_notification *get_tagged_notification(struct mako_state *state, const char *tag, const char *app_name);
 size_t format_notification(struct mako_notification *notif, const char *format,
 	char *buf);
 void notification_handle_button(struct mako_notification *notif, uint32_t button,
--- a/notification.c
+++ b/notification.c
@@ -44,6 +44,7 @@ void reset_notification(struct mako_notification *notif) {
 	free(notif->body);
 	free(notif->category);
 	free(notif->desktop_entry);
+	free(notif->tag);
 	if (notif->image_data != NULL) {
 		free(notif->image_data->data);
 		free(notif->image_data);
@@ -55,6 +56,7 @@ void reset_notification(struct mako_notification *notif) {
 	notif->body = strdup("");
 	notif->category = strdup("");
 	notif->desktop_entry = strdup("");
+	notif->tag = strdup("");
 
 	notif->image_data = NULL;
 
@@ -127,6 +129,17 @@ struct mako_notification *get_notification(struct mako_state *state,
 	struct mako_notification *notif;
 	wl_list_for_each(notif, &state->notifications, link) {
 		if (notif->id == id) {
+			return notif;
+		}
+	}
+	return NULL;
+}
+
+struct mako_notification *get_tagged_notification(struct mako_state *state,
+		const char *tag, const char *app_name) {
+	struct mako_notification *notif;
+	wl_list_for_each(notif, &state->notifications, link) {
+		if (strcmp(notif->tag, tag) == 0 && strcmp(notif->app_name, app_name) == 0) {
 			return notif;
 		}
 	}


emersion:
Thanks for the PR! Is there a spec for this somewhere? Who uses it?

Would be nice to open a MR against https://gitlab.freedesktop.org/xdg/xdg-specs to add it.

williamvds:
This is a private spec from Canonical's NotifyOSD, [documented here](https://wiki.ubuntu.com/NotifyOSD). It's intended for "confirmation bubbles" to show a response to an action. A side-effect is the hint being used to replace other notifications that used the same value for the hint.
It's also implemented by Dunst, another notification server, [documented here](https://github.com/dunst-project/dunst/blob/master/docs/dunst.pod).

I've looked for any mention of it amongst the XDG repos and I'm surprised to find nothing. I'll take a look at raising an issue about this in https://gitlab.freedesktop.org/xdg/xdg-specs

Looks like @akobel worked on the same thing with #271, so it's up to you which you take to decide to accept.
I'm happy to add criteria matching support like he has - might as well support it.

akobel:
Thanks, I was not aware of this spec.

To be fair, the full specification by Canonical has some other relevant points for common (read: my ;-)) use cases: those confirmation feedback bubbles should be click-through, and (assuming #36 and/or #38 will be implemented) sounds (and perhaps also exec's) should play only after a fade-in, and unless the notification has been replaced.

Rationale: pushing the XF86AudioLowerVolume with such visual feedback a couple of times in a row should play a ping in the new volume as soon as you stop hitting the key, but not for each an every key press; otherwise, it'd drive you bonkers, in particular if the old chime is not stopped. Neither of our PRs prepares for anything like that.

emersion:
>Personally I prefer updated notifications being bumped up if they're replaced when they're supposed to be sorted by time, and I think replacing by ID should do the same

For the record, the spec specifically says implementations should _not_ do this.

emersion:
Also, the whole "synchronous tag" idea is a little fragile. It's easy to get tag collisions.

williamvds:
> For the record, the spec specifically says implementations should _not_ do this.

Alright, I'll update the patch to ensure behavior meets the spec

> Also, the whole "synchronous tag" idea is a little fragile. It's easy to get tag collisions.

That's a fair issue. Perhaps it could be alleviated by combining it with other data, like the desktop entry?  
For my own scripts using `notify-send` it's kinda on me if I introduce collisions in those tags.

The appeal of the tag is how it provides a simpler API for replacing notifications, particularly when using only `notify-send`. The existing replace-by-ID method is more precise, but it requires active tracking of those IDs.  
For simple things like showing the new volume when I change it, it adds a lot more hassle than I'd like to deal with.

dbus/xdg.c
```diff
@@ -240,6 +240,16 @@ static int handle_notify(sd_bus_message *msg, void *data,
 			// will win over app_icon if provided.
 			free(notif->app_icon);
 			notif->app_icon = strdup(image_path);
+		} else if (strcmp(hint, "synchronous") == 0 ||
+				strcmp(hint, "private-synchronous") == 0 ||
+				strcmp(hint, "x-canonical-private-synchronous") == 0 ||
+				strcmp(hint, "x-dunst-stack-tag") == 0) {
```
emersion, line 246: Hm. Do we really need to support all of these? `private-synchronous` and `x-canonical-private-synchronous` seem a little alien. `synchronous` seems like it's polluting the hint namespace. I get that this isn't standardized, but do we have an idea how much these are used in practice?

include/notification.h
```diff
@@ -57,7 +58,8 @@ enum mako_notification_close_reason {
 	MAKO_NOTIFICATION_CLOSE_EXPIRED = 1,
 	MAKO_NOTIFICATION_CLOSE_DISMISSED = 2,
 	MAKO_NOTIFICATION_CLOSE_REQUEST = 3,
-	MAKO_NOTIFICATION_CLOSE_UNKNOWN = 4,
+	MAKO_NOTIFICATION_CLOSE_REPLACED = 4,
+	MAKO_NOTIFICATION_CLOSE_UNKNOWN = 5,
```
emersion, line 62: We can't do this, because the reason is part of the spec.

"Replacing" a notification should behave (from the user PoV) as if the original notification had been updated, not closed.

notification.c
```diff
@@ -48,6 +48,9 @@ void reset_notification(struct mako_notification *notif) {
 	free(notif->body);
 	free(notif->category);
 	free(notif->desktop_entry);
+	if (notif->tag != NULL) {
```
emersion, line 51: No need to check for NULL here.

emersion:
Sorry for the long hiatus. I guess I'm fine with implementing this feature if regular clients won't use it, even with the drawbacks we discussed about. There is probably some value in supporting existing hint names to allow cross-notification-daemon scripts.

This PR needs to be rebased.

dbus/xdg.c
williamvds: The list was copied from those supported by Dunst; I thought it was apt to spare people migrating over to mako from modifying their existing scripts.

`x-dunst-stack-tag` and `private-synchronous` are used by Dunst alone. `x-canonical-private-synchronous` is the only hint officially supported by NotifyOSD, but `synchronous` is [supported for compatibility with older applications](https://bazaar.launchpad.net/~indicator-applet-developers/notify-osd/trunk/view/head:/src/stack.c#L696).

`private-synchronous` appears to be intended as a notification-daemon independent alternative.

How about going with `x-canonical-private-synchronous` for compatibility with newer applications using Canonical's "official" private API, and `private-synchronous` for general usage by scripts?

williamvds:
Rebased and applied recommended changes. It'll now search and directly replace matching notifications in the list. This saves us the trouble of trying to find the right notification to replace before we've processed the hints.

dbus/xdg.c
emersion: > private-synchronous appears to be intended as a notification daemon agnostic alternative.

It seems like notify-osd was using `private-synchronous` before they were made aware that non-standard keys must be vendored. Ref: https://bugs.launchpad.net/ubuntu/+source/notify-osd/+bug/340817

> How about going with x-canonical-private-synchronous for compatibility with newer applications using Canonical's "official" private API, and private-synchronous as an agnostic alternative for scripts?

I'd rather not recommend `private-synchronous` at all, since it's been removed from the original implementation in notify-osd. Also the "private" prefix doesn't really make sense (it probably means "private to Canonical").

I'm fine with supporting `x-dunst-stack-tag` and `x-canonical-private-synchronous`.

Also ref the dunst PR: https://github.com/dunst-project/dunst/pull/552

emersion: Hm, there's also a difference between the cap name and the hint name: https://github.com/dunst-project/dunst/pull/552#issuecomment-433630749

dbus/xdg.c
```diff
@@ -343,6 +361,16 @@ static int handle_notify(sd_bus_message *msg, void *data,
 	}
 	notif->requested_timeout = requested_timeout;
 
+	if (notif->tag) {
+		// Find and replace the existing notfication with a matching tag
+		struct mako_notification *replace_notif = get_tagged_notification(state, notif->tag);
+		if (replace_notif) {
+			wl_list_insert(&replace_notif->link, &notif->link);
+			destroy_notification(replace_notif);
```
emersion, line 369: This doesn't replace the notification, this closes the previous one and opens a new one. This has consequences on e.g. ordering and IDs.

See how `replaces_id` is handled at the top of the function.

emersion: tl;dr we need to reset the replaced notification, and then to copy over all the fields from `notif` to the replaced notification.

emersion: Maybe an easier trick would be to override `notif->id` with the replaced notification's, then `wl_list_insert(&replaced->link, &notif->link)` to insert the "new" notif at the right position, then destroy `replaced`.

dbus/xdg.c
williamvds: Adjusted to only support `x-canonical-private-synchronous` and `x-dunst-stack-tag`. Kept the capabilities as they are to `private-synchronous` and `x-canonical-private-synchronous` since it's difficult to tell exactly what's expected. Seems reasonable to me to go with what Canonical supports and most applications are using.

dbus/xdg.c
williamvds: Updated to override the new `notif->id`. It means we lose an ID, but it saves the hassle of copying the new notification into the original.

dbus/xdg.c
emersion: It seems like the capabilities match the hint names, or at least dunst assumes so. Can we remove `private-synchronous` from the caps, and add `x-dunst-stack-tag`?

dbus/xdg.c
emersion: Yup, that sounds reasonable!

include/notification.h
```diff
@@ -60,7 +61,7 @@ enum mako_notification_close_reason {
 	MAKO_NOTIFICATION_CLOSE_EXPIRED = 1,
 	MAKO_NOTIFICATION_CLOSE_DISMISSED = 2,
 	MAKO_NOTIFICATION_CLOSE_REQUEST = 3,
-	MAKO_NOTIFICATION_CLOSE_UNKNOWN = 4,
+	MAKO_NOTIFICATION_CLOSE_UNKNOWN = 5,
```
emersion, line 64: This should be 4, not 5.

notification.c
```diff
@@ -133,6 +135,17 @@ struct mako_notification *get_notification(struct mako_state *state,
 	return NULL;
 }
 
+struct mako_notification *get_tagged_notification(struct mako_state *state,
+		const char *tag) {
+	struct mako_notification *notif;
+	wl_list_for_each(notif, &state->notifications, link) {
+		if (notif->tag && strcmp(notif->tag, tag) == 0) {
```
emersion, line 142: Should we try to match by `app_id` as well? ie. the tags are namespaced per-app?

It doesn't seem like dunst does that, so maybe we shouldn't. If we were to standardize this capability, I think this would be desirable to avoid tag collisions.

emersion:
As a side note, it would be nice to submit a proposal to standardize this. One possible pain point is that this capability isn't meant to be used for long-running processes, I think its main use-case is scripts. Properly explaining that regular notification clients shouldn't use it may be tricky.

The upstream repo is https://gitlab.freedesktop.org/xdg/xdg-specs.

notification.c
williamvds: I think it makes sense too. Most scripts probably won't bother setting the app ID anyway, but every little helps. I'll add it for now.

emersion (approved):
LGTM, thanks!

notification.c
```diff
@@ -133,6 +135,17 @@ struct mako_notification *get_notification(struct mako_state *state,
 	return NULL;
 }
 
+struct mako_notification *get_tagged_notification(struct mako_state *state,
+		const char *tag, const char *app_name) {
+	struct mako_notification *notif;
+	wl_list_for_each(notif, &state->notifications, link) {
+		if (strcmp(notif->tag, tag) == 0 && strcmp(notif->app_name, app_name) == 0) {
```
williamvds, line 142: Just noticed a glaring issue with this check: notifications that don't provide any tag at all are going to match each other. Will open a fix.

williamvds:
Thanks for the reviews!

I've finally opened an issue to make this an official hint in the XDG specification: https://gitlab.freedesktop.org/xdg/xdg-specs/-/issues/77

emersion:
Excellent, thanks!

PR #271, opened 2020-05-06
Handle synchronous group matches for notification replacements

akobel:
Gosh! Looking at #270, @williamvds beat me to it. So, looks like there's more support for having this integrated... ;-)

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -31,6 +31,7 @@ void destroy_criteria(struct mako_criteria *criteria) {
 	free(criteria->app_icon);
 	free(criteria->category);
 	free(criteria->desktop_entry);
+	free(criteria->synchronous_group);
 	free(criteria->summary);
 	free(criteria->body);
 	free(criteria->raw_string);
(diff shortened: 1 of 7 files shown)


akobel:
That said: @williamvds has a cleaner and nicer diff. I +1 his choice of `tag` rather than my clumsy `synchronous_group` field name, even though it's very generic.
I jump through some hoops to keep the id intact, but probably that doesn't actually matter... At least dunst just increments; and since synchronous is not a core of the spec, AFAICS, it should be implementation-defined behavior anyway...

It'd be somewhat nice to be able to use the `tag`/`synchronous_group` in criteria, even though probably `app-name` suffices.

So, no objections if his version is taken... ;-)

akobel:
Potential problem with both our approaches: the new notification (meant as a replacement) is put at the end of the list rather than inserted where it was. Hence, makoctl dismiss targets this notification as the topmost one on the stack, and it's considered as a new entry for the history...

emersion:
Superseded by https://github.com/emersion/mako/pull/270

PR #275, opened 2020-05-29
Add regex support for summary criteria

TheAvidDev:
This patch adds POSIX extended regular expression matching support for summary criteria.

I'm new to C and mako, so let me know anything that needs to be fixed / changed, but so far the following are parts I'd like to clarify:
```c
regexec(&criteria->summary, notif->summary, 0, NULL, 0)
```
This could error out, so should there be a check and output to stderr?

---
```diff
-	criteria->summary = strdup(notif->summary);
```
I'm not sure if this is okay or not. The `summary` for a notification's `mako_criteria` isn't currently accessed anywhere and the `mako_notification` already has the notification's summary, so I think it's alright. While it could be filled in by compiling the summary to regex, I don't think there's any point in doing so.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -280,19 +280,19 @@ bool apply_criteria_field(struct mako_criteria *criteria, char *token) {
 		} else if (strcmp(key, "summary") == 0) {
 			criteria->summary = strdup(value);
 			if (criteria->spec.summary_pattern) {
-				fprintf(stderr, "Cannot set both summary and summary-pattern.");
+				fprintf(stderr, "Cannot set both summary and summary~ regex.\n");
 				return false;
 			}
 			criteria->spec.summary = true;
 			return true;
-		} else if (strcmp(key, "summary-pattern") == 0) {
+		} else if (strcmp(key, "summary~") == 0) {
 			if (regcomp(&criteria->summary_pattern, value,
 					REG_EXTENDED | REG_NOSUB)) {
-				fprintf(stderr, "Invalid summary-pattern regex '%s'\n", value);
+				fprintf(stderr, "Invalid summary~ regex '%s'\n", value);
 				return false;
 			}
 			if (criteria->spec.summary) {
-				fprintf(stderr, "Cannot set both summary and summary-pattern.");
+				fprintf(stderr, "Cannot set both summary and summary~ regex.\n");
 				return false;
 			}
 			criteria->spec.summary_pattern = true;
--- a/criteria.c
+++ b/criteria.c
@@ -32,7 +32,8 @@ void destroy_criteria(struct mako_criteria *criteria) {
 	free(criteria->app_icon);
 	free(criteria->category);
 	free(criteria->desktop_entry);
-	regfree(&criteria->summary);
+	free(criteria->summary);
+	regfree(&criteria->summary_pattern);
 	free(criteria->body);
 	free(criteria->raw_string);
 	free(criteria);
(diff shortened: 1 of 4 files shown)


criteria.c
```diff
@@ -368,7 +371,6 @@ struct mako_criteria *create_criteria_from_notification(
 	criteria->urgency = notif->urgency;
 	criteria->category = strdup(notif->category);
 	criteria->desktop_entry = strdup(notif->desktop_entry);
-	criteria->summary = strdup(notif->summary);
```
vilhalmer, line 371: This actually is used, when determining how to group notifications (see dbus/xdg.c:395, for example). Currently even with this change grouping will still require exact matches, since there's no way in the group option to specify a regex for this field. I think that's fine for the initial implementation, we may not ever want to get fancier than that.

Anyway, this is going to require creating a regex that matches exactly the literal summary string of the notification. I don't see an included function for escaping regex special characters in regex(3), which is unfortunate. We will either have to write our own, which needs to escape all of `.^$*+?()[{\|`, or hack around the problem by having a separate field and a flag that determines whether to compare using the regex field or the string field. Not really a fan of the hack.

criteria.c
```diff
@@ -82,7 +83,7 @@ bool match_criteria(struct mako_criteria *criteria,
 	}
 
 	if (spec.summary &&
-			strcmp(criteria->summary, notif->summary) != 0) {
+			regexec(&criteria->summary, notif->summary, 0, NULL, 0) == REG_NOMATCH) {
```
vilhalmer, line 86: I think your option of spitting an error out potential errors on stderr is fine at least for the first pass. We don't really have a more robust error reporting mechanism in place, and most users will be keeping an eye on stderr while creating new configuration anyway.

vilhalmer:
A potential path forward that would cause keeping both a regex and string summary in criteria to _not_ be a hack would be to add a separate summary-pattern criteria field and only allow the user to provide one or the other in the config file. This would keep the current behavior compatible and simple for most users, and users who want to get fancy can use the new field. `create_criteria_from_notification` would then always use the existing field.

TheAvidDev:
> ... add a separate summary-pattern criteria field and only allow the user to provide one or the other in the config file. 

I agree, this is the best option to go forward with. I'll try to get it done soon.

TheAvidDev:
I've switched to the `summary-pattern` criteria @vilhalmer, but I suspect some of my styling may be off, so let me know if that's the case.

emersion:
Code LGTM, but I wonder if we can come up with a better syntax. If we're going to support regexes for other fields too, maybe adding support for something like `[summary~=regex]` would make sense? (syntax up to discussion)

vilhalmer:
Ah, I like that idea! Do we want to do that as part of this PR or have a followup?

emersion:
I'm fine with a follow-up PR, just need to make the final decision before next release to avoid breaking the configuration format.

vilhalmer:
Sounds good to me. @TheAvidDev if you want to take that on in this PR you're more than welcome, if not I can open an issue to track that and we can merge this as-is.

TheAvidDev:
For my own use case, I only really need the `summary-pattern` but I agree that adding this to other fields is a generally good idea (and a pretty cool feature I haven't seen any other notification daemon implement). I'm not actually sure how long this will take me to do so I think having it as a separate PR be better.

If we for sure want this for other fields, perhaps I can just change `summary-pattern` to `summary~` which should keep the config part the same for the future, even if the code eventually changes to allow other fields?

emersion:
>If we for sure want this for other fields, perhaps I can just change summary-pattern to summary~ which should keep the config part the same for the future, even if the code eventually changes to allow other fields?

Ah, that's an interesting idea, and probably a good idea to mitigate config syntax breakage. :)

Anyone has opinions about the color of the `~=` bikeshed? Any better syntax ideas?

vilhalmer:
Very clever, I'm all for doing it this way since it gets us the syntax effectively for free. I think the tilde is a fine choice.

TheAvidDev:
> Any better syntax ideas?

While I think something like `criteria-pattern` or `criteria-regex` is clearer, they are also pretty verbose. I kinda like the simplicity of the `~=` and it might make it easier to parse.

So +1 for `criteria~=pattern`.

emersion:
All right, nice, we all agree! Can you update the PR and the docs, and then we can merge this?

TheAvidDev:
Done.

vilhalmer:
Thanks!

PR #276, opened 2020-06-02
Fix more memory issues

williamvds:
- acbbeaa: `state.config` needs to be freed before exiting
- f6c9f0d: `components` needs to be freed before returning
- acbbeaa: old values of `config_arg` should be freed in case the `-c` option is specified multiple times, and `config_arg` needs to be freed if returning before `load_config_arg()` is called
- f77b6e5: bit iffy about this one: scan-build thinks it's possible for a notification in the history list to remain in the list after being destroyed, which would cause a use-after-free and/or double-free. I'm quite sure that's impossible, but I rewrote the history pruning to use a for each loop instead and that's made the warning disappear

### DIFF
--- a/config.c
+++ b/config.c
@@ -756,18 +756,27 @@ int parse_config_arguments(struct mako_config *config, int argc, char **argv) {
 
 	optind = 1;
 	char *config_arg = NULL;
+	int opt_status = 0;
 	while (1) {
 		int option_index = -1;
 		int c = getopt_long(argc, argv, "hc:", long_options, &option_index);
 		if (c < 0) {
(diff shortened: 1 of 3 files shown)


emersion:
>Fix memory leaks caused by -c option

Nit: would prefer something like `opt_status` instead of `short_status`.

>scan-build thinks it's possible for a notification in the history list to remain in the list after being destroyed, which would cause a use-after-free and/or double-free

I don't think that's possible. Please leave this commit out, it makes the code logic more complicated.

williamvds:
Done and done!

PR #279, opened 2020-06-21
Improve robustness of config integer parsing

### LINKED ISSUE #130: Negative value of max-visible
Hi!

Starting playing with Mako I'm not sure if I'm geeting all of the wording beind the `max-visible` things:

> If -1, all notifications are visible

When I set this to `-X` and then execute something like `notify-send "foo"` mako display the notification and an extra "(X+1 more)" as shown in the following screenshot:

Untitled

Not sur if its unwanted or I didn't get the meaning of this option.

Thanks for the contribution with Mako!

vilhalmer:
Instead of parsing directly into the target config attribute, use a temporary variable. This avoids cases where we could report parsing as failed, like for '57asdf', but would have already set the value to 57. Also validate that integers are in range, for options which have one.

Fixes #130, and a bunch that haven't been reported yet. :)

### DIFF
--- a/config.c
+++ b/config.c
@@ -485,9 +485,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 	} else if (strcmp(name, "text-color") == 0) {
 		return spec->colors.text = parse_color(value, &style->colors.text);
 	} else if (strcmp(name, "width") == 0) {
-		return spec->width = parse_int(value, &style->width);
+		return spec->width = parse_int_ge(value, &style->width, 1);
 	} else if (strcmp(name, "height") == 0) {
-		return spec->height = parse_int(value, &style->height);
+		return spec->height = parse_int_ge(value, &style->height, 1);
 	} else if (strcmp(name, "margin") == 0) {
 		return spec->margin = parse_directional(value, &style->margin);
 	} else if (strcmp(name, "padding") == 0) {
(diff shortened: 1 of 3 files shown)


vilhalmer:
Should wait until #228 goes in so I can rebase on that and do the necessary rearranging.

Ongy:
What exactly canges in the "57asdf" example with this?
To my understanding it will error out either way.

And looking at the code, I feel like a `parse_int_min` or `parse_int_ge` might make sense. All those comparisons are of the form `>=` and giving that a name is usually nice.

vilhalmer:
`strtol` will happily parse 57 out of that string and assign it, we add the extra check on top of it. Since we're assigning the result of strtol directly to the out pointer in `parse_int`, it can get set even if we then decide the string was invalid.

vilhalmer:
(This isn't an issue at startup since parse errors kill mako, but it could be a problem when reloading the config.)

Ongy:
Does it actually? The way I see it we would still end up in https://github.com/emersion/mako/blob/master/config.c#L817 and refuse the config to be parsed.

And I was wondering if that change can't be put into `parse_int`. The only other use-site is in criteria.c https://github.com/emersion/mako/blob/master/criteria.c#L274 and at least wouldn't be hurt by this.

vilhalmer:
Ah, you're right; I forgot we do it atomically. I'll merge this all into the parser as you suggest.

emersion:
Needs rebase

vilhalmer:
Done!

Issue #283, opened 2020-06-27
Better support for actions


shroff:
#195 does a decent job of allowing interaction with notification actions via a keybinding to `makoctl menu`, but it's still not quite there for me yet.

I want to be able to interact with any notification (not just the latest one), and also use the cursor to select actions because that is a lot more intuitive for people that I share my computer with.

I have a working solution that doesn't require major modifications or any additional dependencies. I'd be happy to send a PR if the approach seems viable.

Here is my approach:
1. Allow adding custom actions to `on-button-*` and `on-touch`
  a. Add a substitution parameter `%i` that is replaced with the notification id.
2. Allow `makoctl menu` to accept an `id` arg
3. Allow `makoctl dismiss` to accept an `id` arg

This allows a mako config like the following:
```
on-button-left=invoke-custom-action
custom-action=sh -c 'makoctl menu -n %i wofi -d -p "Select Action..." && makoctl dismiss %i'
```

The biggest downside is that it allows for arbitrary command execution on notification click...

emersion:
This PR implements some of this: https://github.com/emersion/mako/pull/249

>Allow makoctl menu to accept an id arg  
>Allow makoctl dismiss to accept an id arg

Yes. We already support this for `makoctl invoke`, so it'd be nice to also do it for those two.

emersion:
Since the only remaining part of this is https://github.com/emersion/mako/issues/38, closing.

PR #287, opened 2020-07-14
Accept id args in makoctl for menu and dismiss

shroff:
Follow-up from the short discussion in #283.

There are things around `dismiss` that are not ideal, but I didn't want to introduce breaking changes without first discussing it. Ideally, we would do the following:
1. `makoctl dismiss` would look like `makoctl dismiss [-n id] [-g]`, and `makoctl dismiss-all` would be a separate action.
2. `DismissLastNotification` dbus action is redundant, but can remain around for compatibility

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -70,7 +70,8 @@ static int handle_dismiss_notification(sd_bus_message *msg, void *data,
 	struct mako_state *state = data;
 
 	uint32_t id = 0;
-	int ret = sd_bus_message_read(msg, "u", &id);
+	uint32_t dismiss_group = 0;
+	int ret = sd_bus_message_read(msg, "ub", &id, &dismiss_group);
 	if (ret < 0) {
 		return ret;
 	}
@@ -86,7 +87,11 @@ static int handle_dismiss_notification(sd_bus_message *msg, void *data,
 	struct mako_notification *notif;
 	wl_list_for_each(notif, &state->notifications, link) {
 		if (notif->id == id) {
-			close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
+			if (dismiss_group) {
+				close_group_notifications(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
+			} else {
+				close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
+			}
 			set_dirty(notif->surface);
 			break;
 		}
@@ -351,7 +356,7 @@ static const sd_bus_vtable service_vtable[] = {
 	SD_BUS_METHOD("DismissAllNotifications", "", "", handle_dismiss_all_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("DismissGroupNotifications", "", "", handle_dismiss_group_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("DismissLastNotification", "", "", handle_dismiss_last_notification, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("DismissNotification", "u", "", handle_dismiss_notification, SD_BUS_VTABLE_UNPRIVILEGED),
+	SD_BUS_METHOD("DismissNotification", "ub", "", handle_dismiss_notification, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("InvokeAction", "us", "", handle_invoke_action, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("RestoreNotification", "", "", handle_restore_action, SD_BUS_VTABLE_UNPRIVILEGED),
 	SD_BUS_METHOD("ListNotifications", "", "aa{sv}", handle_list_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -70,7 +70,7 @@ static int handle_dismiss_notification(sd_bus_message *msg, void *data,
 	struct mako_state *state = data;
 
 	uint32_t id = 0;
-	uint32_t dismiss_group = 0;
+	int dismiss_group = 0;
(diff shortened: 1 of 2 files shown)


makoctl
```diff
@@ -73,19 +75,28 @@ case "$1" in
...
-			call InvokeAction "us" 0 "$sel"
-		fi
+	if [ $# -gt 1 ] && [ "$1" = "-n" ]; then
+		id=$2
+		actions="$(call ListNotifications | id=$id jq --arg id $id -re '.data[0][] | select(.id.data==($id | tonumber)) | .actions.data')"
+		shift 2
 	else
+		actions="$(call ListNotifications | jq -re '.data[0][0].actions.data')"
+		id="0"
+		echo $id
```
emersion, line 88: Remnant from debugging logs?

makoctl
```diff
@@ -35,20 +37,21 @@ fi
...
 	"-g"|"--group")
 		call DismissGroupNotifications
 		;;
-	"")
-		call DismissLastNotification
-		;;
 	*)
-		echo "makoctl: unrecognized option '$2'"
-		exit 1
+		re='^[0-9]+$'
```
emersion, line 49: I don't think it's worth it to try to validate the ID, `busctl` will do it anyway. For instance:

```
Failed to parse 'asdf' as unsigned 32bit integer: Invalid argument
```

emersion:
For consistency with other sub-commands, can we add a `-n <id>` option to `dismiss` instead of having it as an argument?

shroff:
> For consistency with other sub-commands, can we add a `-n <id>` option to `dismiss` instead of having it as an argument?
True, though the semantics of the dismiss sub-command are still inconsistent because the `-a` option doesn't quite make sense with `-n` or `-g`, which play nice with each other.

I'm happy to do it that way, but IMO it would make sense to make `makoctl dismiss-all` a separate sub-command albeit at the expense of compatibility.

Thoughts?

emersion:
>it would make sense to make makoctl dismiss-all a separate sub-command albeit at the expense of compatibility.

What's the reasoning behind this?

shroff:
> > it would make sense to make makoctl dismiss-all a separate sub-command albeit at the expense of compatibility.
> 
> What's the reasoning behind this?

The options aren't independently composable, i.e. `-n <id>` and `-g` are basically ignored with `-a`

emersion:
I'd still prefer to stick with the current option naming convention.

makoctl
```diff
@@ -73,19 +94,27 @@ case "$1" in
...
-		sel="$(jq -rn --arg sel "$sel" "$array|to_entries[]|select(.value == \$sel).key")"
-		if [ -z "$sel" ]; then
-			echo >&2 "$0: No action selected"
-			exit 1
-		else
-			call InvokeAction "us" 0 "$sel"
-		fi
+	if [ $# -gt 1 ] && [ "$1" = "-n" ]; then
+		id=$2
+		actions="$(call ListNotifications | id=$id jq --arg id $id -re '.data[0][] | select(.id.data==($id | tonumber)) | .actions.data')"
```
emersion, line 99: `id=$id` is unnecessary

emersion: `--arg id $id` should probably be quoted: `--arg id "$id"`

makoctl
```diff
@@ -73,19 +94,27 @@ case "$1" in
...
-		sel="$(jq -rn "$array|values[]" | "$@")"
-		sel="$(jq -rn --arg sel "$sel" "$array|to_entries[]|select(.value == \$sel).key")"
-		if [ -z "$sel" ]; then
-			echo >&2 "$0: No action selected"
-			exit 1
-		else
-			call InvokeAction "us" 0 "$sel"
-		fi
+	if [ $# -gt 1 ] && [ "$1" = "-n" ]; then
+		id=$2
```
emersion, line 98: `$2` should be quoted here

makoctl
```diff
@@ -33,22 +37,39 @@ fi
 
 case "$1" in
 "dismiss")
-	[ $# -lt 2 ] && action="" || action="$2"
-	case "$action" in
-	"-a"|"--all")
+	all=0
+	group=0
+	id=0
+	while [ $# -gt 1 ]; do
```
emersion, line 43: Why do we need a loop here? Wasn't the previous code enough?

makoctl
```diff
@@ -33,22 +37,39 @@ fi
...
+		*)
+			echo >&2 "$0: Unrecognized option: $2"
+			exit 1
+			;;
+		esac
+		shift
+	done
+
+	if [ $all -eq 1 ]; then
+		echo "all!!"
```
emersion, line 68: Remnant from debugging session?

dbus/mako.c
```diff
@@ -65,6 +65,42 @@ static int handle_dismiss_last_notification(sd_bus_message *msg, void *data,
 	return sd_bus_reply_method_return(msg, "");
 }
 
+static int handle_dismiss_notification(sd_bus_message *msg, void *data,
+		sd_bus_error *ret_error) {
+	struct mako_state *state = data;
+
+	uint32_t id = 0;
+	uint32_t dismiss_group = 0;
```
emersion, line 73: Per the table in `sd_bus_message_read(3)`, this should be an `int`.

dbus/mako.c
```diff
@@ -65,6 +65,42 @@ static int handle_dismiss_last_notification(sd_bus_message *msg, void *data,
...
+	int ret = sd_bus_message_read(msg, "ub", &id, &dismiss_group);
+	if (ret < 0) {
+		return ret;
+	}
+
+	if (id == 0) {
+		id = state->last_id;
+	}
+
+	if (wl_list_empty(&state->notifications)) {
```
emersion, line 83: Nit: technically this `if` is unnecessary, `wl_list_for_each` is a no-op on empty lists.

makoctl
shroff: So that you can use `-g` with `-n <id>`

dbus/mako.c
shroff: That's the pattern in `handle_invoke_action` and I assumed it was there for a good reason. Removed from here.

makoctl
emersion: Ah, makes sense

PR #289, opened 2020-07-23
Add dockerfiles for building and packaging on debian

niklashhh:
Added crude dockerfiles for
1. Compiling mako from source
1. Creating a crude debian package (not for distributing)

Added instructions for using the dockerfiles.

I hope they can be helpful with ci too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ Mkfile.old
 dkms.conf
 
 build/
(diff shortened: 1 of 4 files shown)


emersion:
Thanks for your PR, but I'm not interested in using Docker, nor adding packaging for a specific Linux distribution.

PR #291, opened 2020-07-31
Fix regression in dismissing multiple notifications

maximbaz:
Follow-up to https://github.com/emersion/mako/commit/b7c3a3ba9849712be920c038ab0b0432c6c45cd8 that supports the following scenario:

```
$ notify-send 1
$ notify-send 2
$ makoctl dismiss
$ makoctl dismiss
```

With this PR, both notifications will be dismissed; on master only one gets dismissed.

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -76,13 +76,9 @@ static int handle_dismiss_notification(sd_bus_message *msg, void *data,
 		return ret;
 	}
 
-	if (id == 0) {
-		id = state->last_id;
-	}
-
 	struct mako_notification *notif;
 	wl_list_for_each(notif, &state->notifications, link) {
-		if (notif->id == id) {
+		if (notif->id == id || id == 0) {
 			if (dismiss_group) {
 				close_group_notifications(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
 			} else {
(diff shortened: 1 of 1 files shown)


emersion:
Hmm. A zero ID should dismiss the last notification. Is there something wrong in the D-Bus handler?

https://github.com/emersion/mako/blob/b7c3a3ba9849712be920c038ab0b0432c6c45cd8/dbus/mako.c#L80

maximbaz:
Oh, and I thought it's a simple typo 🙂 

So the first time `last_id` is correct, but I imagine after dismissal the value of `last_id` doesn't get updated.

I see it is changed only in one place (incremented upon creating a new notification), does it mean that on dismiss the value should be decremented? And if a group of notifications is closed, then should be decremented by a number of them?

https://github.com/emersion/mako/blob/81edd54f8479385f79d5c4da34f7be2a90271b38/notification.c#L77

emersion:
Ah, interesting. I'd prefer not to decrement it, instead directly picking the last list item.

maximbaz:
OK I didnt notice at first that there is conveniently a list of all notifications available 🙂 This works too, I'm only not sure if this is the best way to retrieve the first element of the list or there are simpler ways?

vilhalmer:
I believe you can use `wl_container_of(notif->state->notifications.next, tmp, link)` (not tested, might be slightly wrong).

maximbaz:
You are right, thanks

maximbaz:
I agree with the comments, could you please point me to some examples of how to implement the alternative solution, how to extract the last notification from D-Bus?

emersion:
In practice that would mean remove all `last_id` usage in `dbus/mako.c`: [here](https://github.com/emersion/mako/blob/b7c3a3ba9849712be920c038ab0b0432c6c45cd8/dbus/mako.c#L79) and [here](https://github.com/emersion/mako/blob/b7c3a3ba9849712be920c038ab0b0432c6c45cd8/dbus/mako.c#L111), replacing it with the logic to get the first element in the list (also used [here](https://github.com/emersion/mako/blob/b7c3a3ba9849712be920c038ab0b0432c6c45cd8/dbus/mako.c#L40) for instance).

maximbaz:
Applied your suggestion, how does the code look now? 🙂

PR #293, opened 2020-08-03
systemd: improve ExecCondition to support absence of $WAYLAND_DISPLAY

maximbaz:
Following up from #240, the current `ExecCondition` relies on the environment variable `WAYLAND_DISPLAY` to be set, but this in turn relies on the fact that... sway somehow sets it for systemd to see (only when sway is started by systemd?).

Regardless of the specifics, referring to the [infamous thread](https://github.com/swaywm/sway/issues/5160) about systemd service for sway, there are examples when users aren't using `environment.d` (I'm one of them). But even though I don't run sway via systemd, I thought I could benefit from mako being auto-started via DBUS-activation of this systemd unit.

This PR changes the `ExecCondition` so that even if systemd doesn't see my `WAYLAND_DISPLAY` variable, it can still detect the presence of the wayland session.

Thoughts?

cc @WhyNotHugo

### DIFF
--- a/contrib/systemd/mako.service.in
+++ b/contrib/systemd/mako.service.in
@@ -7,7 +7,7 @@ After=graphical-session.target
 [Service]
 Type=dbus
 BusName=org.freedesktop.Notifications
-ExecCondition=/bin/sh -c '[ -n "$WAYLAND_DISPLAY" ]'
+ExecCondition=/bin/sh -c '[ -e "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}" ]'
 ExecStart=@bindir@/mako
 ExecReload=@bindir@/makoctl reload
 


emersion:
I don't think this is the right solution. Maybe the user is running a test Sway session inside their main X11 session.

maximbaz:
🤔 Do I understand correctly that the current systemd unit is only meant to work when sway is running via systemd as well? Or how is sway supposed to set $WAYLAND_DISPLAY so that this unit sees this variable?

WhyNotHugo:
This solution does expect you to be running sway as a systemd service (though it's compatible with that).

If you're not running it as a systemd service (it's not recommended to do so anyway), you can export your sway environment to system by running this:

    systemctl --user import-environment; systemctl --user start sway-session.target

I have this near the end of my sway config file. It's explain further [in sway's wiki](https://github.com/swaywm/sway/wiki/Systemd-integration#managing-user-applications-with-systemd).

maximbaz:
Thanks for the answers! 

> Maybe the user is running a test Sway session inside their main X11 session.

In this case, won't dunst already be running, and so D-BUS will not try to activate mako?

> You can export your sway environment to system by running `systemctl --user import-environment`

Won't this be prone to the problem that @emersion mentioned? Suppose you run X11, you run sway in a test session, `$WAYLAND_DISPLAY` gets imported into systemd environment, then you close sway but the systemd environment still contains that environment variable?

Or maybe taking another example, if you have say Wayland on tty1 and X11 on tty2, once you log into sway on tty1, you import `WAYLAND_DISPLAY` into systemd environment, then you exit sway, switch to X11 on tty2, but systemd still contains `WAYLAND_DISPLAY`?

If we check for the file, then the file would at least disappear once we exit sway, while if we rely on environment variable, then users are expected to load and unload the variable as they enter and exit sway.

Or maybe I'm misunderstanding something? 🙂 

(btw new systemd release contains `ConditionEnvironment`, but it's still in testing on Arch)

emersion:
>Suppose you run X11, you run sway in a test session

Yeah, you should only import env into systemd for your main session, not for test/secondary sessions.

>If we check for the file

There's no guarantee the file will be named wayland-0 btw.

The only way to check whether you're in a Wayland session is to check for `WAYLAND_DISPLAY`. This doesn't play well with systemd and dbus, I don't think there's a perfect solution.

If mako is started without the right `WAYLAND_DISPLAY` variable set, maybe it'll be started on the wrong compositor.

maximbaz:
Good points, I can't think of anything better too... Let's just close this then, starting mako process directly from sway is at least predictable and cannot go wrong, as opposed to messing with systemd environment or trying some file detections.

Thanks for your input!

WhyNotHugo:
The existing condition is already a hack to avoid breakage when people meet all of these conditions:

- Run X11
- Have mako installed
- Rely on D-Bus activation  to start a notification server.

It's already an unusual enough scenario. If you regularly switch between Xorg and Wayland, DBus activation isn't for you, and you'll need to explicitly eager-start you notification daemon.

You can't have multiple desktop ecosystems running (o run without rebooting) AND use DBus activation. It's just a recipe for things to break.

maximbaz:
That I agree with, but in that case why do we setup any condition at all, if it is only to catch a scenario that is such specific anyway? Dunst is not doing any conditions to ensure that X11 is running and Wayland isn't.

PR #299, opened 2020-09-27
Update AppArmor profile

foxcpp:
Not sure why it actually works with `dbus-strict` but in principle it should be `dbus-session-strict` since dbus-strict is for system bus, not a session one.


emersion (approved):
Thanks!

PR #301, opened 2020-10-18
Add surface criteria and make max-visible and hidden actually work per surface

vilhalmer:
This is the result of a couple weeks of mulling over discussion from IRC about whether `max-visible` and `hidden` can be made to work as intended with implicit surfaces. TL;DR yes, but I'm marking this draft because it needs some feedback on whether this is actually the right approach, and if I missed any cases.

I added new criteria fields for `output` and `anchor`. The combination of these is effectively the same as giving surfaces names, because they are the two unique attribues of any surface. So now you can do this:
```
[output=DP-6 anchor=top-right]
background-color=#ffffff
max-visible=2
```
These are not magically applied to surfaces instead of notifications, though. They behave much like `grouped` and `group-index` already did: the first pass of criteria will potentially set them, and the second pass then re-styles the notifications based on their final values. It's still only two passes, so it's not possible to get into an infinite loop.

Treating them as normal styles allows you to do interesting things like I've done above, changing the appearance one last time for anything that appears on a given surface. It's a little weird to have `max-visible` stored on every notification rather than the surface, but it has the same effect since they all just happen to get the same value set.

On its own the above is insufficient, as it would allow specifying criteria that don't make sense:
```
[anchor=top-right]
group-by=anchor
```
This cannot possibly work, since we only do two passes. I noticed we already had some situations like this. For example at the moment you can specify `group-by` inside a criteria maching `grouped` which would just silently do nothing. I've made all of these impossible cases fail the config validation, and tried to give them useful messages as well.

TODO:
* ~Make it invalid to use `max_visible` inside criteria matching anything other than `anchor` and `output`. While you might be able to do clever things specifying it on other matches, it violates the principle of least astonishment as it would be applied in config order rather than by the smallest max-visible value. (In fact it's impossible to accomplish the latter even with a special case, due to the presence of the default value.)~ Done.
* ~Fix up `hidden` so you can style that per surface as well.~ Done.
* ~Much more testing.~ I've been running this since october and haven't seen any crashes or weird behavior.
* ~Finish adding new criteria to the manpage.~ Done.

Let me know what you think of this solution!

### DIFF
--- a/config.c
+++ b/config.c
@@ -733,7 +733,7 @@ int load_config_file(struct mako_config *config, char *config_arg) {
 
 	// Validate the final criteria section since there was no opening bracket
 	// after it to do this in the loop.
-	if (criteria != NULL && !validate_criteria(criteria)) {
+	if (ret != -1 && criteria != NULL && !validate_criteria(criteria)) {
 		fprintf(stderr, "Invalid configuration in criteria: [%s]\n",
 				criteria->raw_string);
 		ret = -1;
--- a/config.c
+++ b/config.c
@@ -691,6 +691,15 @@ int load_config_file(struct mako_config *config, char *config_arg) {
 		}
 
 		if (line[0] == '[' && line[strlen(line) - 1] == ']') {
+			// Since we hit the end of the previous criteria section, validate
+			// that it doesn't break any rules before moving on.
+			if (criteria != NULL && !validate_criteria(criteria)) {
+				fprintf(stderr, "Invalid configuration in criteria: [%s]\n",
+						criteria->raw_string);
+				ret = -1;
+				break;
+			}
+
 			free(section);
 			section = strndup(line + 1, strlen(line) - 2);
 			if (strcmp(section, "hidden") == 0) {
@@ -742,6 +751,14 @@ int load_config_file(struct mako_config *config, char *config_arg) {
 			ret = -1;
 			break;
 		}
+	}
+
+	// Validate the final criteria section since there was no opening bracket
+	// after it to do this in the loop.
+	if (criteria != NULL && !validate_criteria(criteria)) {
+		fprintf(stderr, "Invalid configuration in criteria: [%s]\n",
+				criteria->raw_string);
+		ret = -1;
 	}
 
 	free(section);
--- a/config.c
+++ b/config.c
@@ -579,29 +579,7 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 		style->spec.layer = true;
 		return true;
 	} else if (strcmp(name, "anchor") == 0) {
-		if (strcmp(value, "top-right") == 0) {
-			style->anchor = ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP |
-				ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
-		} else if (strcmp(value, "top-center") == 0) {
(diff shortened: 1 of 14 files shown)


vilhalmer:
One issue I've noticed: the first frame isn't output-styled. Might need to push an empty first frame to let all the info settle after a new layer is created.

vilhalmer:
Trying to configure `max-visible` more granularly than by surface will now result in the following:
```
…
[output=DP-7 anchor=top-right app-name=asdf]
max-visible=2
```
```
Setting `max_visible` is allowed only for `output` and/or `anchor`
Invalid configuration in criteria: [output=DP-7 anchor=top-right app-name=asdf]
Failed to parse config
```

vilhalmer:
Still need to update the manpage, but this should be about ready to go now.

emersion:
Hm, can you explain the issues that this PR is trying to fix? I've paged that out of my brain, unfortunately.

vilhalmer:
There are two issues:

- `[hidden]` didn't allow styling the placeholder differently for each surface
-  `max-visible` was a global option, which meant you couldn't configure it differently per surface

Both are fixed in the same way, by adding criteria for selecting the output and anchor that define the surface. These are second-pass criteria, similar to the grouping ones, since you need to figure out which output/anchor a notification is on before you can match those values. Hidden is now a proper criteria so it can be specified in the same match as these, like:
`[anchor=top-left hidden]`

config.c
```diff
@@ -691,6 +691,15 @@ int load_config_file(struct mako_config *config, char *config_arg) {
 		}
 
 		if (line[0] == '[' && line[strlen(line) - 1] == ']') {
+			// Since we hit the end of the previous criteria section, validate
+			// that it doesn't break any rules before moving on.
+			if (criteria != NULL && !validate_criteria(criteria)) {
```
emersion, line 678: This doesn't validate the last criteria section.

Maybe we can just loop over all criteria sections after parsing them?

criteria.c
```diff
@@ -456,3 +448,34 @@ struct mako_criteria *create_criteria_from_notification(
...
+		if (criteria->style.spec.anchor) {
+			fprintf(stderr, message, "anchor");
+			return false;
+		} else if (criteria->style.spec.output) {
+			fprintf(stderr, message, "output");
+			return false;
+		} else if (criteria->style.spec.group_criteria_spec) {
+			fprintf(stderr, message, "group-by");
+			return false;
+		}
```
emersion, line 507: Maybe this would be slightly better?

```c
const char *invalid_field = NULL;
if (criteria->style.spec.anchor) {
    invalid_field = "anchor";
} else if (…) {
    …
}
if (invalid_field) {
    fprintf(stderr, "Setting …", invalid_field);
    return false;
}
```

Compilers don't usually like it when the format string is stored in a variable.

config.c
emersion: Ah, this is fixed in a subsequent commit. Would be nice to squash these two.

vilhalmer: Good idea, will do.

criteria.c
vilhalmer: Yep, I like that a lot better.

vilhalmer:
Updated the manpage.

emersion (approved):
I haven't done an in-depth review, but overall this looks good to me! I think at this point I'd rather merge this now and handle any potential regression/bug afterwards, to stop the Great Mako Release Deadlock.

Thanks a lot for working on this!

PR #305, opened 2020-10-27
Add support for body criteria

### LINKED ISSUE #304: Notification criteria for message body?
Hi!

I'm loving mako so far, and just now looking into its criteria options. It looks like it has a bunch of filters, but weirdly no filter for message body text.

Is this a technical limitation, or could this be added?

My use case is to filter useless notifications from a particular application, [Solaar](https://github.com/pwr-Solaar/Solaar/). None of the other fields really offer any distinction - for the same device, it will use the same summary, app-name, app-icon, category and desktop-entry. The only distinction between different notifications is the notification body text. They've [decided that filtering useless from useful notifications is the job of the notification manager](https://github.com/pwr-Solaar/Solaar/issues/335#issuecomment-612442973), so ideally I'd do this in mako's config.

daboross:
This adds the ability to match notifications based on notification bodies, either using an exact string match or an approximate regex. See issue #304.

I've based this almost entirely on the existing summary support - so, it should be at least as functional as that! Notably, there was also existing scaffolding for matching body criteria in the codebase. It was missing from config parsing, though, and from the documentation. Since the TODO in the original code mentioning body support described it as waiting for regex support, I've also added regex support for bodies, matching the existing regex support for summaries.

This was a fairly straightforward PR - the only real thing I did was pull out regex criteria matching, to avoid duplication between summaries and bodies - everything else is just directly copied and modified from the existing summary code.

Let me know if there's anything I can do. And I hope this is useful!

Fixes #304.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -36,8 +36,23 @@ void destroy_criteria(struct mako_criteria *criteria) {
 	free(criteria->summary);
 	regfree(&criteria->summary_pattern);
 	free(criteria->body);
+	regfree(&criteria->body_pattern);
 	free(criteria->raw_string);
 	free(criteria);
+}
+
+static bool match_regex_criteria(regex_t *pattern, char *value) {
+	int ret = regexec(pattern, value, 0, NULL, 0);
+	if (ret != 0) {
+		if (ret != REG_NOMATCH) {
+			size_t errlen = regerror(ret, pattern, NULL, 0);
+			char errbuf[errlen];
+			regerror(ret, pattern, errbuf, sizeof(errbuf));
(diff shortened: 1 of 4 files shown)


TheAvidDev (approved):
LGTM.

As a follow up, and as discussed in #275, we want regex support for all fields by using `~=` instead of `=`. I think doing this in a separate PR is fine since this one is primarily about adding body criteria support.

emersion (approved):
LGTM as well, thanks!

>we want regex support for all fields by using ~= instead of =. I think doing this in a separate PR is fine since this one is primarily about adding body criteria support.

Agreed. Filled #306 for this.

PR #309, opened 2020-10-31
replaced echo's with cat, replaced tests with case, prefixed left-par…

GH-TpaeFawzen:
…enthesis to patterns.

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -1,28 +1,30 @@
 #!/bin/sh -eu
 
 usage() {
-	echo "Usage: makoctl <command> [options...]"
-	echo ""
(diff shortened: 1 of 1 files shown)


emersion:
Why?

GH-TpaeFawzen:
> Why?

replacing echo's with cat: to reduce number of commands to be called.
replacing some test commands with case statement: test command is not always built-in while case statement is.
left-parenthesis: POSIX lets you do so.

emersion:
Thanks for sending a pull request, but I'm not interested in micro-optimizations. Readability is more important.

PR #311, opened 2020-11-09
Add support for basu

emersion:
See https://github.com/emersion/basu

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -38,10 +38,16 @@ if get_option('sd-bus-provider') == 'auto'
 	if not sdbus.found()
 		sdbus = dependency('libelogind',
 			required: false,
+			not_found_message: 'libelogind not found, trying basu',
 		)
 	endif
 	if not sdbus.found()
-		error('Neither libsystemd, nor libelogind was found')
+		sdbus = dependency('basu',
+			required: false,
+		)
+	endif
+	if not sdbus.found()
+		error('Neither libsystemd, nor libelogind, nor basu was found')
 	endif
 else
 	sdbus = dependency(get_option('sd-bus-provider'))
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
 	'c',
 	version: '1.4.1',
 	license: 'MIT',
-	meson_version: '>=0.47.0',
+	meson_version: '>=0.50.0',
 	default_options: [
 		'c_std=c11',
 		'warning_level=2',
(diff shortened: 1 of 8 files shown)


meson.build
```diff
@@ -29,13 +29,30 @@ realtime = cc.find_library('rt')
...
+			not_found_message: 'libelogind not found, trying basu',
+		)
+	endif
+	if not sdbus.found()
+		sdbus = dependency('basu',
+			required: false,
+		)
+	endif
+	if not sdbus.found()
+		error('Nor libsystemd, nor libelogind, nor basu was found')
```
vilhalmer, line 50: Should start with "None of" or "Neither". However, would it be better to mark basu required, and use its not_found_message?

vilhalmer (approved):
Looks good to me, just one comment on the failure mode if none are found.

meson.build
emersion: Oh right, fixed.

> However, would it be better to mark basu required, and use its not_found_message?

I was trying to not make the error message specific to a single potential sd-bus provider (since we try them all).

vilhalmer: I was thinking you could use this existing error message to replace the default not_found_message and then wouldn't need a final if block. I don't remember if the format that meson prints it out in would look good, though. Either way is fine with me.

emersion: Looks like this:

```
Run-time dependency basu found: NO (tried pkgconfig and cmake)
Message: Hey there

../meson.build:29:0: ERROR: Dependency "basu" not found, tried pkgconfig and cmake
```

I'm not a fan of the message being in-between two Meson messages which reference "basu".

Here's what happens with the `error` approach:

```
Run-time dependency basu found: NO (tried pkgconfig and cmake)

../meson.build:30:0: ERROR: Problem encountered: Hey there
```

vilhalmer: Yep, agreed the current way is better. :+1:

TheAvidDev (approved):
LGTM!

kchibisov:
It seems like `makoctl` should be updated, since right now it uses `busctl`, however basu provides `basuctl`, thus it should be adjusted somehow, though, I'm not sure what will be the policy to package basu.

emersion:
Indeed, that's a good point. We could just select between `busctl` and `basuctl` at runtime.

emersion:
>We could just select between busctl and basuctl at runtime.

Fixed

PR #313, opened 2020-11-21
Add missing flags to makoctl manpage

vilhalmer:
This brings the manpage into alignment with the help output. I also rephrased all instances of "first notification" to "a notification" and specified the default on the `-n` flag, and added a hint about where to get action names.

### DIFF
--- a/makoctl.1.scd
+++ b/makoctl.1.scd
@@ -14,8 +14,8 @@ Sends IPC commands to the running mako daemon via dbus.
 
 # COMMANDS
 
-*dismiss*
-	Dismisses the first notification.
+*dismiss* [-a|--all] [-g|--group] [-n <id>]
+	Dismisses a notification.
 
 	Options:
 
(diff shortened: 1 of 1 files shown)


makoctl.1.scd
```diff
@@ -25,26 +25,43 @@ Sends IPC commands to the running mako daemon via dbus.
...
-	Invokes an action on the first notification. If _action_ is not specified,
-	invokes the default action.
+*invoke* [-n <id>] [action]
+	Invokes an action on a notification. If _action_ is not specified, invokes
+	the default action. Action names can be discovered using `makoctl list`.
+
+	Options:
+
+	*-n* <id>
+	    Invoke the action on the notification with the given id. Defaults to
```
emersion, line 42: This is using spaces instead of tabs, and makes the build fail I believe.

vilhalmer: Indeed, I wonder what plugin failed to only use spaces in those three spots… sigh.

emersion (approved):
Thanks!

PR #320, opened 2020-11-29
mako.5.scd: fix typo

pschervakov:
border-size: default=1 -> default=2

https://github.com/emersion/mako/blob/5f5e43f94a292c43f20ea7b98120a3bc880ec1d6/config.c#L87


emersion (approved):
Thanks

PR #324, opened 2020-12-21
Unbreak build on FreeBSD

jbeich:
Depends on https://github.com/emersion/basu/pull/32

Runtime seems to work fine but I haven't tested much:
- notify-send -> shows popup
- makoctl dismiss -> clears popup

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -1,0 +1,20 @@
+image: freebsd/latest
+packages:
+  - basu
+  - gdk-pixbuf2
+  - libepoll-shim
+  - meson
+  - pango
+  - pkgconf
+  - scdoc
+  - wayland
+  - wayland-protocols
+sources:
+  - https://github.com/emersion/mako
+tasks:
+  - setup: |
+      cd mako
+      meson build/ -Dauto_features=enabled -Dsd-bus-provider=basu -Dsystemd=disabled
+  - build: |
+      cd mako
+      ninja -C build/
--- a/meson.build
+++ b/meson.build
@@ -29,6 +29,12 @@ realtime = cc.find_library('rt')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
 
+epoll = dependency('', required: false)
+if (not cc.has_function('timerfd_create', prefix: '#include <sys/timerfd.h>') or
+    not cc.has_function('signalfd', prefix: '#include <sys/signalfd.h>'))
+	epoll = dependency('epoll-shim')
+endif
+
 if get_option('sd-bus-provider') == 'auto'
 	assert(get_option('auto_features').auto(), 'sd-bus-provider must not be set to auto since auto_features != auto')
 	sdbus = dependency('libsystemd',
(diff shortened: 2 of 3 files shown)


meson.build
```diff
@@ -21,6 +21,7 @@ mako_inc = include_directories('include')
 cc = meson.get_compiler('c')
 
 cairo = dependency('cairo')
+epoll = dependency('epoll-shim', required: false)
```
emersion, line 24: Can we do something like wlroots here?

```meson
if host_machine.system() == 'freebsd'
	libepoll = dependency('epoll-shim')
else
	libepoll = dependency('', required: false)
endif
```

This avoids the check on non-FreeBSD, and fails the build if missing on FreeBSD, making it clearer it's not an optional dependency.

.builds/freebsd.yml
```diff
@@ -0,0 +1,20 @@
+image: freebsd/latest
+packages:
+  - basu
```
kennylevinsen, line 3: I assume we'll have to build it ourselves for now, [like we do on alpine](https://github.com/emersion/mako/blob/master/.builds/alpine.yml).

emersion: Could use a subproject if it's simpler, but last time I tried basu wouldn't build as a subproject.

meson.build
jbeich: epoll-shim is required on DragonFly, FreeBSD, NetBSD, OpenBSD and probably forks like MirBSD, MidnightBSD, GNU/kFreeBSD. However, only DragonFly and FreeBSD currently package/support wlroots compositors. DragonFly would need to port basu before noticing missing epoll-shim dependency in mako.

Some of those may follow IllumOS lead by moving epoll emulation into the kernel. I'd prefer to avoid platform whitelists which quickly grow out of date.

https://illumos.org/man/5/epoll
http://src.illumos.org/source/search?defs=epoll&project=illumos-gate

emersion: I agree platform checks aren't great. Maybe replace with something like `cc.has_function('epoll')`?

meson.build
```diff
@@ -29,6 +30,10 @@ realtime = cc.find_library('rt')
 wayland_client = dependency('wayland-client')
 wayland_protos = dependency('wayland-protocols', version: '>=1.14')
 
+if not cc.check_header('sys/timerfd.h', dependencies: epoll)
```
emersion, line 33: Maybe check for `sys/epoll.h` instead? `dependencies: epoll` is probably unnecessary.

Is `cc.ahs_function('epoll')` not working?

emersion: Ah, or are you checking for `sys/timerfd.h` because that's the only reason why we need `epoll`? Makes sense then.

We should probably migrate to a simple signal handler that writes to a pipe instead, but that's out of scope for this PR.

.builds/freebsd.yml
jbeich: If basu API is stable there's not much point *not* to use system package.

emersion: Yup, if it's available using the system package is better indeed. I wasn't sure whether it would just be in ports or also in pkg repos, and it now seems like it's the latter. Thanks for pushing it forward!

emersion (approved):
Thanks!

jbeich:
Now that basu is tested on FreeBSD would it make sense to switch Alpine to `-Dsd-bus-provider=elogind`?

emersion:
Indeed, would be better to get full coverage for all providers!

PR #325, opened 2020-12-26
ci: Test elogind on Alpine

jbeich:
After #324 basu is already tested on FreeBSD. Let's cover elogind on Alpine instead.


emersion (commented):
Can we also remove the `basu` task, as well as the basu deps and sources?

emersion (approved):
LGTM, thanks!

PR #331, opened 2021-01-17
Fix manpage formatting

deviant:
scdoc requires explicit line breaks. Without, this renders as:

`Default: <b>%s</b>\n%b Default when grouped: (%g) <b>%s</b>\n%b`

(Note: I'm not entirely sure whether this solution is correct: I didn't
apply indentation to the second line, as otherwise it is indented twice.
This might be a bug in scdoc, or could be its intended behaviour. Either
way, this was surprising to me, so I thought I'd mention just in case.)


emersion (approved):
Thanks for the fix!

PR #338, opened 2021-02-15
contrib: drop apparmor and systemd build-system integration

emersion:
We get too many distribution-specific bug reports about them.

Let's not install these by default. Let's leave it to distributions to adapt and ship these.

### DIFF
--- a/contrib/apparmor/meson.build
+++ b/contrib/apparmor/meson.build
@@ -1,8 +1,0 @@
-if get_option('apparmor')
-	apparmordir = join_paths(get_option('sysconfdir'), 'apparmor.d')
-	install_data(
-		'fr.emersion.Mako',
-		install_dir: apparmordir,
-		install_mode: 'rw-r-----',
-	)
-endif
--- a/contrib/systemd/mako.service
+++ b/contrib/systemd/mako.service
@@ -9,8 +9,8 @@ ConditionEnvironment=WAYLAND_DISPLAY
 [Service]
 Type=dbus
 BusName=org.freedesktop.Notifications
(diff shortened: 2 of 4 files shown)


vilhalmer (approved):
Looks extremely good to me. :)

WhyNotHugo:
The [D-Bus service](https://github.com/emersion/mako/blob/bb7e8e4/fr.emersion.mako.service.in) file is now broken, and should not be installed any more either (since it references the no-longer-installed-by-default systemd file).

emersion:
Fixed in https://github.com/emersion/mako/commit/dbd9c2bc98ef0cec7801a049f4bf8d61f7be36d0

WhyNotHugo:
dbd9c2b just makes everything worse.

I've put a drop-in `mako.service` file to run mako as a systemd service (this works before dbd9c2b). With dbd9c2b, however, mako is now started as a non-systemd service by the dbus daemon.

There's no way for me to get mako running as a systemd service any more, unless I manually edit `/usr/share/dbus-1/services/fr.emersion.mako.service`. However, users should not be expected to edit a system file this.

emersion:
Distributions should.

WhyNotHugo:
So distributions now need to manually patch this file, and also install the `systemd.service` file, right?

Can we include such a patch for `fr.emersion.mako.service` could be included in `contrib` too? I'm just looking to avoid downstreams from having to each redo the same work (and also prevent so much fragmentation).

emersion:
It's just a single line to be appended to the D-Bus service file. Maybe `contrib/systemd/README` with some explanations would be nice to have.

PR #344, opened 2021-03-10
Add text-alignment option

YerinAlexey:
This option sets text alignment in the notifications like `dunst`. Possible values are: `left`, `center` or `right`


emersion (approved):
LGTM, thanks!

PR #345, opened 2021-03-11
Simplify evdev include on FreeBSD

jbeich:
Inspired by https://github.com/DragonFlyBSD/DeltaPorts/commit/24bdb05a7f54


emersion (approved):
Thanks!

Issue #300, opened 2020-10-08
Colors for scalable icons


sometoby:
Is there any chance the colors of scalable icons could be made customizable, perhaps even with a default of `text-color`?

emersion:
How would that look like from a pango perspective?

sometoby:
I don't know, I'm not familiar with any of those APIs or mako's internals.

PR #347, opened 2021-05-01
Color in svg icons with text-color

robertgzr:
Took a shot at implementing #300. Not sure how to go about detecting if we should load it as a scalable so I match on the extension for now... This might need something more robust though.

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -18,6 +18,7 @@
 #ifdef HAVE_ICONS
 
 #include <gdk-pixbuf/gdk-pixbuf.h>
+#include <gio/gio.h>
 #include "cairo-pixbuf.h"
 
 static bool validate_icon_name(const char* icon_name) {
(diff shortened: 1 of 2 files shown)


icon.c
```diff
@@ -18,14 +18,88 @@
...
+
+static GdkPixbuf *load_image(const char *path, struct mako_style *style) {
 	if (strlen(path) == 0) {
 		return NULL;
 	}
 	GError *err = NULL;
-	GdkPixbuf *pixbuf = gdk_pixbuf_new_from_file(path, &err);
+	GdkPixbuf *pixbuf;
+	// FIXME ok? or do we need to actually look at mime info?
+	if (strcmp(strrchr(path, '.'), ".svg") == 0) {
```
robertgzr, line 98: not sure about this one... let me know if there is something better we can do? I've been using it with this and haven't had any issues so far.

emersion:
I must say, I'm not a huge fan of mungling SVGs to set the icon colors… It seems like the classes are pretty GNOME-specific too?

robertgzr:
@emersion yeah I basically followed https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gdkpixbufutils.c#L335

I personally don't care about the classes, we can just drop those. `rect,circle,path` are basic svg elements so I don't see an issue with them. 

if we want to allow to match them with our foreground color I don't think there's a way around the munging...

robertgzr:
I actually noticed that svg icons supplied by applications (e.g. mpv) weren't looking right previously...

now it will not overwrite icons that use a color fill already

emersion:
Sorry, I'd rather not add this feature.

octvs:
Is it possible to keep the feature living on a fork @robertgzr ? Analogous to [mortie/swaylock-effects](https://github.com/mortie/swaylock-effects) for [swaywm/swaylock](https://github.com/swaywm/swaylock).

PR #349, opened 2021-05-03
Implement xdg-activation-v1

emersion:
This allows us to pass down an activation token to the invoked
notification. The receiver can then use the token to activate
its toplevel.

The Wayland protocol has been added in [1] and the D-Bus spec
additions are in [2].

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/50
[2]: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/43

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -1,4 +1,4 @@
-image: alpine/latest
+image: alpine/edge
 packages:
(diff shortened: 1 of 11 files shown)


emersion:
This is now ready for review.

PR #351, opened 2021-05-07
Add anchor to center-right and center-left

krtko1:
I was really missing this feature. I was hacking it by creating huge margins, but it somehow stop working.

Tested on sway 1.6 on Arch linux


emersion (approved):
Thanks!

PR #353, opened 2021-05-18
Fix notification replacement tag comparison

williamvds:
Check that the tag is provided at all with strlen() before comparing the
contents of the tag. Otherwise notifications that don't provide any
value for the tag will be matched with and replace each other.

Issue mentioned in #270


emersion (approved):
Good catch, thanks!

PR #354, opened 2021-05-19
Add missing declaration of fish_comp

Kirens:
Compiling with `-Dfish-completions` seems to fail with
```
contrib/completions/meson.build:14:14: ERROR: Unknown variable "fish_comp".
```
I guess @ammgws meant to have a declaration like this. However I have no clue what I'm doing :sweat_smile:, but this seems to compile and install as expected.


emersion (approved):
LGTM, thanks!

PR #359, opened 2021-06-03
Print notification category in `makoctl list`

### LINKED ISSUE #358: Feature Request: Dismiss all notifications with given category
It would be great if the `makoctl dismiss -g` flag had an optional `<group name>` argument, to dismiss all notification in a given group/category.

Use case:
I'm sending a notification whenever a new email comes in, setting `--category=mail`.
When I open my email program, I'd like to automatically dismiss all the pending new-email notifications, doing something like `makoctl dismiss -g mail`, but currently the `-g` flag only works on the most recent group.

I could parse the output of `makoctl list`, but sadly that doesn't expose the group/category names of the notifications, so I'd have to resort to less precise filtering by name. (I could make a separate feature request for that.)

If you agree this would be nice, I'd be happy to make a PR.

RagnarGrootKoerkamp:
Fix #358

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -173,6 +173,12 @@ static int handle_list_notifications(sd_bus_message *msg, void *data,
 			return ret;
 		}
 
+		ret = sd_bus_message_append(reply, "{sv}", "category",
+			"s", notif->category);
+		if (ret < 0) {
+			return ret;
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #360, opened 2021-06-04
Fix memory leak by freeing image before returning in error path

pasqualirb:


emersion (approved):
Thanks.

(Note, I think this should error codepath is only used on allocation failure.)

Issue #335, opened 2021-01-31
Proposal: modes


emersion:
This is a counter-proposal for https://github.com/emersion/mako/issues/138.

Instead of allowing to change arbitrary configuration options, we could introduce modes:

```
[mode=dnd]
invisible=1
```

Modes could be switched at runtime with e.g. `makoctl set-mode dnd`. There is only one mode active at a time.

fluix-dev:
Ooh, I like this proposal, especially with #306. This seems much simpler to both use and implement without confusion w.r.t. merging contexts.

The only issue I imagine is that with #138, you can dynamically generate everything on-the-fly, while this proposal requires hard coding all possible values. **However**, because `makoctl reload` is hot reloading, people can resort to modifying the config and reloading which *I* think is a fine compromise (granted I have no purpose for dynamically generating criteria myself).

vilhalmer:
I think this is the right path forward. It also doesn't preclude adding `set` back later if we have some breakthrough on making it usable.

fluix-dev:
> It also doesn't preclude adding set back later if we have some breakthrough on making it usable.

I think this is a main reason I like this proposal. It has no conflicts with `set`.

emersion:
> There is only one mode active at a time.

Side note: I can see how allowing multiple modes active at the same time could be useful, but I'm not sure it's something we want.

fluix-dev:
I don't think there's a need to do it right now either, and in the future we could have `makoctl add-mode <mode>` and `makoctl del-mode <mode>` or similar.

emersion:
I've implemented this in https://github.com/emersion/mako/pull/362. Please try it out!

maximbaz:
Works well for me, thank you for implementing it!

cherti:
When I try `./makoctl set-mode dnd` I get "Call failed: Transport endpoint is not connected". What could be the cause of that? The mako running is the mako that is built from the branch, the dnd-mode is specified (simply copied from your example to the end of my mako-config), commenting out all other options in the config doesn't resolve the issue.

PR #362, opened 2021-06-15
Introduce modes

emersion:
Closes: https://github.com/emersion/mako/issues/335

### DIFF
--- a/mako.5.scd
+++ b/mako.5.scd
@@ -268,6 +268,10 @@ The following fields are available in criteria:
 - _desktop-entry_ (string)
 - _actionable_ (boolean)
 - _expiring_ (boolean)
+- _mode_ (string)
+	- Only apply style options in this section if the provided mode is
+	  currently enabled. For more information about modes, see the _MODES_
+	  section.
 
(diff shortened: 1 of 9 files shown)


vilhalmer (approved):
This looks awesome!

fluix-dev (approved):
Looks really clean!

mako.5.scd
```diff
@@ -388,6 +392,25 @@ specifiers.
...
 
+# MODES
+
+mako supports applying style options conditionally via modes. A configuration
+section with a _mode_ criteria will only be applied if the current mode
+matches. **makoctl**(1) can be used to change the current mode.
+
+The default mode is named "default".
+
+For example, to hide all notifications if the mode "dnd" (Do Not Disturb) is
```
fluix-dev, line 403: Whoops, forgot to add this comment. I'm just a *tad* worried that "Do Not Disturb" makes it seem like this is a dedicated mode when it's just an arbitrary name. I think removing the clarification or using `"custom"` would be better.

emersion: Renamed the mode to `do-not-disturb`.

PR #364, opened 2021-06-27
New Feature: Change cursor when mouse moves over a notification

Davester47:
Mako is really cool, but I noticed that when I moved my mouse over it the cursor would stay the same as what it was before. If it was an ibeam, it would stay an ibeam, and if it was normal, it would stay normal. I made it change to the cursor that is shown when you hover over a button, since it felt weird clicking on notifications before.

These websites were very helpful when I created this PR:
https://bugaevc.gitbooks.io/writing-wayland-clients/content/beyond-the-black-square/cursors.html
https://doc.qt.io/qt-5/qcursor.html#a-note-for-x11-users

The one flaw in this PR is if a notification has rounded corners, it will change the cursor even where the corners have been rounded off. Clicking in that area still makes the notification go away, though, so I think its fine.

I had to add a dependency to get this to work: wayland-cursor.

### DIFF
--- a/include/mako.h
+++ b/include/mako.h
@@ -3,6 +3,7 @@
 
 #include <stdbool.h>
 #include <wayland-client.h>
+#include <wayland-cursor.h>
 #ifdef HAVE_LIBSYSTEMD
 #include <systemd/sd-bus.h>
 #elif HAVE_LIBELOGIND
@@ -57,6 +58,9 @@ struct mako_state {
 	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wl_list outputs; // mako_output::link
 	struct wl_list seats; // mako_seat::link
+	struct wl_cursor_theme *cursor_theme;
+	const struct wl_cursor_image *cursor_image;
+	struct wl_surface *cursor_surface;
 
 	struct wl_list surfaces; // mako_surface::link
 
--- a/wayland.c
+++ b/wayland.c
@@ -142,6 +142,17 @@ static void touch_handle_up(void *data, struct wl_touch *wl_touch,
 
 }
 
+static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface, wl_fixed_t x, wl_fixed_t y) {
+	struct mako_seat *seat = data;
+	struct mako_state *state = seat->state;
+	// Change the mouse cursor to "left_ptr"
+	if (state->cursor_theme != NULL) {
+		wl_pointer_set_cursor(wl_pointer, serial, state->cursor_surface,
+			state->cursor_image->hotspot_x, state->cursor_image->hotspot_y);
+	}
+}
+
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct mako_seat *seat = data;
@@ -168,7 +179,7 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
 }
 
 static const struct wl_pointer_listener pointer_listener = {
-	.enter = noop,
+	.enter = pointer_handle_enter,
 	.leave = noop,
 	.motion = pointer_handle_motion,
 	.button = pointer_handle_button,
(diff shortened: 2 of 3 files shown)


Davester47:
I realized I forgot to destroy the wl_surface I made for the cursor icon when the program closes. This latest commit adds a call to wl_surface_destroy for that.

wayland.c
```diff
@@ -400,6 +411,27 @@ bool init_wayland(struct mako_state *state) {
...
 	}
 
+	// Set up the cursor. It needs a wl_surface with the cursor loaded into it.
+	// If one of these fail, mako will work fine without the cursor being able to change.
+	state->cursor_theme = wl_cursor_theme_load(NULL, 24, state->shm);
+	if (state->cursor_theme == NULL) {
+		fprintf(stderr, "couldn't find a cursor theme\n");
+		return true;
+	}
+	struct wl_cursor *cursor = wl_cursor_theme_get_cursor(state->cursor_theme, "pointing_hand");
```
emersion, line 421: `pointing_hand` is usually used for links, not for buttons or other clickable UI elements. I'd prefer to use `left_ptr` here.

wayland.c
```diff
@@ -400,6 +411,27 @@ bool init_wayland(struct mako_state *state) {
 		}
 	}
 
+	// Set up the cursor. It needs a wl_surface with the cursor loaded into it.
+	// If one of these fail, mako will work fine without the cursor being able to change.
+	state->cursor_theme = wl_cursor_theme_load(NULL, 24, state->shm);
```
emersion, line 416: Use `XCURSOR_SIZE` from the env to allow users to override the cursor size. See https://github.com/emersion/slurp/blob/e77e1c432bcbd101fd084448bab2df173154a0dc/main.c#L889

emersion:
Looks good to me apart from these comments! Can you also squash all of the commits into a single one to keep a clean Git history?

wayland.c
fluix-dev: Isn't `left_ptr` the default pointer so this change wouldn't do anything in this case?

Davester47: @fluix-dev Yes, left_ptr is the default cursor. However, if you move from a text box like the Firefox address bar to a notification, currently you won't get left_ptr, and you'll instead get ibeam.

Davester47:
I added the changes that you suggested, and I was wondering if I should also add a check for XCURSOR_THEME as well. I think I shouldn't, because it will still fetch the theme stored at /usr/share/icons/default. Firefox also seems to ignore the XCURSOR_THEME variable.

wayland.c
emersion: Also happens when a notification pops up over a terminal.

wayland.c
```diff
@@ -142,6 +142,17 @@ static void touch_handle_up(void *data, struct wl_touch *wl_touch,
 
 }
 
+static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface, wl_fixed_t x, wl_fixed_t y) {
+	struct mako_seat *seat = data;
+	struct mako_state *state = seat->state;
+	// Change the mouse cursor to "pointing_hand"
```
emersion, line 149: This comment is outdated

wayland.c
```diff
@@ -400,6 +411,39 @@ bool init_wayland(struct mako_state *state) {
 		}
 	}
 
+	// Set up the cursor. It needs a wl_surface with the cursor loaded into it.
+	// If one of these fail, mako will work fine without the cursor being able to change.
+	const char* cursor_size_env = getenv("XCURSOR_SIZE");
```
emersion, line 416: Style: `*` on the right

wayland.c
```diff
@@ -400,6 +411,39 @@ bool init_wayland(struct mako_state *state) {
...
+	int cursor_size = 24;
+	if (cursor_size_env != NULL) {
+		errno = 0;
+		char* end;
```
emersion, line 420: Ditto

include/mako.h
```diff
@@ -57,6 +58,9 @@ struct mako_state {
 	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wl_list outputs; // mako_output::link
 	struct wl_list seats; // mako_seat::link
+	struct wl_cursor_theme *cursor_theme;
+	struct wl_cursor_image *cursor_image;
```
emersion, line 62: Maybe make this `const`, to make it clearer we don't need to free it?

Davester47:
I corrected the errors that you pointed out. Is there anything else you would like me to fix?

emersion (approved):
LGTM, thanks!

PR #365, opened 2021-07-04
Make bindings style options

emersion:
Allow defining bindings in criteria sections.


vilhalmer (approved):
LGTM! I can't think of a need to collect these in `apply_superset_style` at the moment. I don't think we would gain anything by conditionally attaching the listeners?

emersion:
Yeah, I've wondered about this as well, and thought it wasn't worth the trouble.

Thanks for the review, @vilhalmer!

PR #366, opened 2021-07-05
Add exec and on-notify bindings

emersion:
_See individual commits._

Closes: https://github.com/emersion/mako/issues/38

### DIFF
--- a/mako.5.scd
+++ b/mako.5.scd
@@ -31,9 +31,9 @@ Empty lines and lines that begin with # are ignored.
 
 # BINDING OPTIONS
 
-Specify the action to take per button or touch event. Supported values are
-_none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_ and
-_exec <command>_.
+Bindings allow to perform an action when an event is triggered. Supported
+values are _none_, _dismiss_, _dismiss-all_, _dismiss-group_,
+_invoke-default-action_ and _exec <command>_.
 
 *on-button-left*=_action_
 	Default: invoke-default-action
@@ -47,12 +47,21 @@ _exec <command>_.
 *on-touch*=_action_
 	Default: dismiss
 
+*on-notify* = _action_
+	Default: none
+
 When _exec_ is used, the command will be executed in a POSIX shell. The
 shell variable _id_ will be set to the notification ID. For example, the
 following option will display an interactive action menu on middle click:
 
 ```
 on-button-middle=exec makoctl menu -n "$id" dmenu -p 'Select action: '
+```
+
+The following option will play a sound when a new notification is opened:
+
+```
+on-notify=exec mpv /usr/share/sounds/freedesktop/stereo/message.oga
 ```
 
 # STYLE OPTIONS
--- a/mako.5.scd
+++ b/mako.5.scd
@@ -32,8 +32,8 @@ Empty lines and lines that begin with # are ignored.
 # BINDING OPTIONS
 
 Specify the action to take per button or touch event. Supported values are
-_none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_.
-
+_none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_ and
+_exec <command>_.
 
 *on-button-left*=_action_
 	Default: invoke-default-action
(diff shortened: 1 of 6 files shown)


emersion:
Note: I've went with the variable name `id` to make it feel like a local, temporary variable. I'm not too happy about the variable leaking to all child processes, OTOH it's not _that_ bad. To fix it, we could pass the ID as an argument to sh, and set the variable from inside the shell script, with something like:

```
sh -c 'id="$1"; <command>' mako-exec <id>
```

But this sounds a bit annoying to implement.

mako.5.scd
```diff
@@ -47,6 +47,23 @@ _none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_.
...
+
+When _exec_ is used, the command will be executed in a POSIX shell. The
+shell variable _id_ will be set to the notification ID. For example, the
+following option will display an interactive action menu on middle click:
+
+```
+on-button-middle=exec makoctl menu -n "$id" dmenu -p 'Select action: '
+```
+
+For example, the following option will play a sound when a new notification
```
vilhalmer, line 61: This second "for example" reads a bit weird since there was one above, can probably just remove it.

mako.5.scd
```diff
@@ -31,9 +31,9 @@ Empty lines and lines that begin with # are ignored.
 
 # BINDING OPTIONS
 
-Specify the action to take per button or touch event. Supported values are
-_none_, _dismiss_, _dismiss-all_, _dismiss-group_, _invoke-default-action_.
-
+Bindings allow to execute an action when an event is triggered. Supported
```
vilhalmer, line 34: Maybe "allow performing an action" instead of "execute" to avoid conflating with the exec action?

notification.c
```diff
@@ -352,15 +357,46 @@ void notification_execute_binding(struct mako_notification *notif,
...
+		} else if (pid == 0) {
+			// Double-fork to avoid SIGCHLD issues
+			pid = fork();
+			if (pid < 0) {
+				perror("fork failed");
+				_exit(1);
+			} else if (pid == 0) {
+				char id_str[32];
+				snprintf(id_str, sizeof(id_str), "%" PRIu32, notif->id);
+				setenv("id", id_str, 1);
```
vilhalmer, line 387: I don't feel too strongly either way, but other tools I use that have similar functionality usually namespace the environment variables (like QUTE_WHATEVER in qutebrowser). Would MAKO_ID or something be better just to make it clear where it's coming from since we can't avoid it leaking into child processes?

notification.c
```diff
@@ -352,15 +357,46 @@ void notification_execute_binding(struct mako_notification *notif,
...
+
+				char *const argv[] = { "sh", "-c", binding->command, NULL };
+				execvp("sh", argv);
+				perror("exec failed");
```
vilhalmer, line 391: Should probably reset errno first just to be safe since there's no other error check like with the forks.

vilhalmer (changes requested):
Looks good, just a couple manpage nits and minor things.

notification.c
emersion: Right. I've been wondering about this above: https://github.com/emersion/mako/pull/366#issuecomment-874190215

I've updated the PR to not use env variables, so it's not leaked to child processes anymore. The variable is now purely local to the shell script. Is that enough to remove your concern about namespacing?

notification.c
emersion: From the exec man page:

> The exec() functions return only if an error has occurred.  The return value is -1, and errno is set to indicate the error.

mako.5.scd
emersion: Good idea

mako.5.scd
emersion: Done

notification.c
vilhalmer: Ah, duh.

notification.c
vilhalmer: Yep, I like this better.

vilhalmer (approved):
LGTM!

emersion:
Ah, and one more note: access to other notification data such as title, body and other fields is not provided for now, I figured people can always query it from the ID via `makoctl list` if they need to.

emersion:
Thanks for the review!

PR #370, opened 2021-07-22
build: find scdoc is build system instead of target

sgn:
For supporting cross compilation.
See: https://github.com/void-linux/void-packages/commit/b07214b837b4bba0fc6ad3e1ea226fcd5335ec6d

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,10 +130,10 @@ configure_file(
 	install_dir: datadir + '/dbus-1/services',
 )
 
-scdoc = dependency('scdoc', required: get_option('man-pages'), version: '>= 1.9.7')
+scdoc = dependency('scdoc', required: get_option('man-pages'), version: '>= 1.9.7', native: true)
 
 if scdoc.found()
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -130,7 +130,7 @@ configure_file(
 	install_dir: datadir + '/dbus-1/services',
 )
 
-scdoc = dependency('scdoc', required: get_option('man-pages'), version: '>= 1.9.7')
+scdoc = dependency('scdoc', required: get_option('man-pages'), version: '>= 1.9.7', native: true)
 
 if scdoc.found()
 	sh = find_program('sh')
```
emersion, line 136: We should also set `native: true` here. And in `protocol/meson.build` as well.

emersion (approved):
LGTM, thanks!

meson.build
sgn: Yeah, that's true. These places was `find_program` so I missed. Changed!

PR #376, opened 2021-08-18
Include <errno.h> to use errno in wayland.c

quantum5:
Without the include, using `errno` results in compile error in some environments.


emersion:
Thanks!

PR #379, opened 2021-08-24
RFC: Fix leaks of mako_notification in handle_notify()

pasqualirb:
Unify clean up on error with goto statements.

Is freeing the expected behavior?


pasqualirb:
Also, the commit 4dc382ecc7585d729723 ("Allow create_notification outside handle_notify") introduced `strdup()` calls inside `reset_notification()`, which is called by `destroy_notification()` just before `notif` is freed.

PR #382, opened 2021-09-16
Feature get-mode

toffan:
Hi all,

Add a `get-mode` command to makoctl to retrieve the current mode and a `toggle-mode` command to go back and forth between a mode and default (Eg. to (de)activate a 'do-not-disturb' mode).

Personally these are 2 commands I would use, but I got no clue about how the project is used in other environment.

EDIT: accordingly to the discussion below the `toggle-mode` feature has been discarded and will come in its own PR.

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -302,6 +302,13 @@ static void reapply_config(struct mako_state *state) {
 	}
 }
 
+static int handle_get_mode(sd_bus_message *msg, void* data,
+		sd_bus_error *ret_error) {
+	struct mako_state* state = data;
+
+	return sd_bus_reply_method_return(msg, "s", state->current_mode);
+}
+
 static int handle_set_mode(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct mako_state *state = data;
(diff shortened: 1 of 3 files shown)


vilhalmer:
Thanks for the PR! I'm +1 on `get-mode`.  For `toggle-mode`, I'm less convinced that this needs to be implemented in mako itself when the caller could pretty easily use `get-mode` to figure out which one is the opposite.

It also sort of paints us into a corner if we ever want to have multiple modes active. I think at the very least it should instead be implemented as `cycle-mode [mode1] [mode2]` to make it explicit that only one of them will be active. That way if we do end up supporting multiple active at a time, `toggle` is still available as a verb for turning them on and off without affecting other active modes. As a bonus, this means you could cycle between any two modes instead of assuming the other is default, and we could later support any number of modes in the list (though not sure if this would ever be useful).

asahaf:
@vilhalmer `toggle-mode` could be between the last two used modes, in case we have multiple them.

toffan:
I'm not against refactoring `toggle-mode` into `cycle-mode`, however I'd like to suggest something else.

Let's assume multiple modes can be active at the same time, and let's say one has 3 modes A, B and C that are active. Then it makes sense to toggle one mode at a time to (de)activate it (here `toggle-mode B` would switch between `A+C` and `A+B+C`.

Now, let's be in the current *1 mode at a time* scenario. The only thing you can do is `toggle-mode B` to switch between `B` and `default`.

The issue is, if `A` is active when you `toggle-mode B`. In the first scenario you switch between `A` and `A+B`, when in the second you go from `A` to `B` and then switch between `default` and `B`.

So, what about calling that illegal for now. Instead of going from `A` to `B` just raise an error saying something like "Only one mode can be active at a time, deactivate mode A first (`makoctl set-mode default`)".  It would 1. make it explicit that only one mode can be active at a time 2. leave the door open to extend this feature in the future.

What do you think about it? Waiting for a consensus on this issue, should I only keep `get-mode` on this MR?

Thank you all for your time.

fluix-dev:
> Waiting for a consensus on this issue, should I only keep `get-mode` on this MR?

Yeah, I think that would be good.

toffan:
PR updated: `toggle-mode` has been discarded and will come back in its own PR.

boredland:
Hi there! Is there anything currently blocking this PR?

emersion:
I'm not sure how this command should behave with multiple modes active.

boredland:
I suppose [jq](https://github.com/emersion/mako/pull/382/files#diff-7576801664f739274077c005e528f4eed0c60ac7ad55cd40bdf89cfe943c211bR139) would return a multi-line string with all modes.

emersion:
Alternative implementation: https://github.com/emersion/mako/pull/433

emersion:
Thanks for your PR, but #433 has superseded this patch.

PR #393, opened 2021-11-23
No blur

### LINKED ISSUE #387: Blurry notification/Hi-DPI
Using mako i got very strange behaviour: all the time i receive notifications, first notification gets blurry, but reloading sway/using slurp/getting another notification and all the notifications are no longer blurry.
20211006_04h18m49s_grim
20211006_04h19m00s_grim

I'm using a Hi-DPI laptop, and of course with `output eDP-1 scale 1` everything works fine.

lilydjwg:
This should fix #387.

### DIFF
--- a/include/mako.h
+++ b/include/mako.h
@@ -75,6 +75,7 @@ struct mako_state {
 	struct wl_list notifications; // mako_notification::link
 	struct wl_list history; // mako_notification::link
 	struct wl_array current_modes; // char *
+	uint32_t last_scale;
 
 	int argc;
 	char **argv;
--- a/main.c
+++ b/main.c
@@ -76,6 +76,7 @@ static bool init(struct mako_state *state) {
 	wl_array_init(&state->current_modes);
 	const char *mode = "default";
 	set_modes(state, &mode, 1);
+	state->last_scale = 1;
 	return true;
 }
 
--- a/wayland.c
+++ b/wayland.c
@@ -621,9 +621,10 @@ static void send_frame(struct mako_surface *surface) {
 		return;
 	}
 
-	int scale = 1;
+	int scale = state->last_scale;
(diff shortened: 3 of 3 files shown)


lilydjwg:
The CI failure seems to be unrelated, complaining a build option I didn't touch.

WhyNotHugo (approved):
Works for me. Thanks!

kabouzeid:
Works for me too, thanks for the PR! Hoping this will get merged soon :)

jplatte:
Ping @emersion

WhyNotHugo:
The issue has been less problematic lately; the first notification is only blurry for an instant and then it quickly unblurs. I think this has been [partially/mostly] fixed.

This PR has conflicts now tho.

lilydjwg:
I've removed all other commits and rebased the de-flickering commit on master.

emersion:
This commit doesn't make sense to me. It works fine when there is only one screen, but breaks as soon as there are multiple.

lilydjwg:
Oops. I'll try to fix it when I get access to my other screen.

emersion:
I don't think it's possible to fix without a Sway change to send the `wl_surface.enter` event before the layer-surface is mapped.

lilydjwg:
Hi, I rebased my commit and tested it on my setup. Yes it would be nicer if we could receive an enter enter before the surface is mapped, but I think this is a good-enough workaround for me.

I have two monitors, one is the laptop's internal screen and the other is an external 4K monitor. I mostly work on the external one while I put some informational windows on the laptop's one, so for the most of time notifications will appear on the external one, i.e. it doesn't switch between multiple outputs.

The issue I have is that, the first frame of mako is not only blurry but also has a different size (it seems to be accumulated difference of text height). This patch makes it not "jump" when it re-appears on the same output.

PR #394, opened 2021-11-26
build: use list for sdbus dep, drop sh for scdoc

emersion:


meson.build
```diff
@@ -40,24 +40,7 @@ endif
...
-	endif
-	if not sdbus.found()
-		sdbus = dependency('basu',
-			required: false,
-		)
-	endif
-	if not sdbus.found()
-		error('Neither libsystemd, nor libelogind, nor basu was found')
-	endif
+	sdbus = dependency(['libsystemd', 'libelogind', 'basu'])
```
vilhalmer, line 43: The meson docs seem to imply this should be multiple arguments rather than a list, but they don't actually provide an example. Does the list have a different behavior, or do they both just work?

vilhalmer (approved):
Just one question about the new dependency feature, everything else looks good.

meson.build
emersion: Hm, good question. They appear to work the same. I switched to the multiple arg approach since that's what the docs use.

(Note, Sway and seatd use a list right now.)

vilhalmer: I found the code in question and it seems like it magically collects variadic args into a list, so a list also happens to work: https://github.com/mesonbuild/meson/commit/4deeb907b613760754c5be20e1af019d76b78515#diff-abffc4d23a79ba9faffdc902d5347cccf4d0eef69ee7503b7ad3c6f86a6fc78eR1496

Not sure if this is actually intentional.

vilhalmer:
Thanks!

PR #395, opened 2021-11-29
Add outer-margin to allow extra margin for first/last notifications

### LINKED ISSUE #225: Border rendering with multiple notifications (regression ?)

### LINKED ISSUE #236: FeatureRequest: first notification position offset
currently the notification is drawn over the pause-button, which makes it impossible to click while the notification is shown.
it would be better to first press pause and then the notification.
I would like to define an offset for the first notification.

temporary fix is: set it to a different anchor point.

ThinkChaos:
Hacked this together because I wanted extra top margin for the first notification!
I installed mako yesterday and this is my first time looking at the code, so please excuse any mistakes. I might've missed some docs or something: I searched for "margin" and added the new "outer-margin" next to each result.

I'm not too attached to the name "outer-margin", chose it because @vilhalmer's [comment](https://github.com/emersion/mako/issues/236#issuecomment-589441738) mentioned it, so feel free to suggest something else.

I also took the liberty of updating the margin docs to include the order for the values since it wasn't clear to me before reading the code.

~~F.i.x.e.s #225: Set outer-margin to compensate negative bottom margin. Example: `margin=0,0,-4,0 outer-margin=0,0,4,0`~~
Fixes #227
Fixes #236


vilhalmer:
Thanks for working on this! Some general comments before I dig in deeper:
- I don't think this actually fixes #225, that needs a determination on whether borders were ever intended to collapse. Fixing it doesn't need to be a goal here either though.
- You don't actually need to do much/any math for this feature, the layer shell protocol can do it for you! We used to use this feature before margins became more flexible, you can probably just restore the code removed in 5397e978. (See also the protocol doc: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml#L190)

I'm also tempted to say that we should implement the collapsing behavior with this margin so the user doesn't have to do math to fix their outer margin when adjusting the inner margins. That may be more complicated than it's worth though. @emersion do you feel strongly one way or another? This could be done in a separate change if we want it.

ThinkChaos:
Thanks for the tip regarding `zwlr_layer_surface_v1_set_margin`. I pushed a commit that reverts my previous changes in render.c and wayland.c and uses that function instead.

I don't feel strongly about #225, so I'll defer to both of you.  
Do note that with the new commit, `outer-margin` can no longer compensate for a negative `margin`. That was personally not a goal of mine, so as I said I'm ok with or without it. I had mentioned it just because I found the issue searching for "margin" and realized the first commit could also serve as a fix.

I can squash my changes if you want before you merge.

ThinkChaos:
Any chance on getting the updated code reviewed?

vilhalmer (approved):
Looks good to me, squashing into a single commit is a good idea. We can adjust collapsing behavior before the next release if we decide it's a good idea.

ThinkChaos:
Not sure what happened, but when I first pushed the squashed commit, GH didn't detect it. So I deleted the branch and pushed again. Sorry for the noise.

vilhalmer:
Thanks! Sorry for the slow process, I haven't had much time lately.

ThinkChaos:
No problem, thanks for getting back to it!

PR #406, opened 2022-01-17
wayland: Support HiDPI for cursor

Nefsen402:


Nefsen402:
!393 Seems to do the same thing, plus some extra.

wayland.c
```diff
@@ -176,6 +176,42 @@ static void touch_handle_up(void *data, struct wl_touch *wl_touch,
 	seat->touch.pts[id].surface = NULL;
 }
 
+void load_cursor(struct mako_state *state, uint32_t scale) {
+	const char *cursor_name = "left_ptr";
+
+	//don't reload the cursor if what we have already can be used
+	if (state->cursor.theme != NULL && state->cursor.scale == scale)
```
emersion, line 183: Style: braces are mandatory. Applies to the whole patch.

wayland.c
```diff
@@ -176,6 +176,42 @@ static void touch_handle_up(void *data, struct wl_touch *wl_touch,
 	seat->touch.pts[id].surface = NULL;
 }
 
+void load_cursor(struct mako_state *state, uint32_t scale) {
```
emersion, line 179: Nit: this function can be `static`.

wayland.c
```diff
@@ -186,10 +222,20 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	seat->pointer.y = wl_fixed_to_int(surface_y);
 	seat->pointer.surface = get_surface(state, wl_surface);
 
+	int scale = 1;
+
+	struct mako_surface *surface;
+	wl_list_for_each(surface, &state->surfaces, link) {
+		if (surface->surface_output->scale > scale) scale = surface->surface_output->scale;
```
emersion, line 229: Style: braces are mandatory

emersion (changes requested):
Thanks for the patch! Here are some minor comments. The meat of the patch looks good.

emersion (approved):
Thanks!

PR #409, opened 2022-01-20
wayland: Cache scale so that first frame respects HiDPI

Nefsen402:
Requires: https://github.com/emersion/mako/pull/410

### DIFF
--- a/wayland.c
+++ b/wayland.c
@@ -621,21 +621,24 @@ static void send_frame(struct mako_surface *surface) {
 		return;
 	}
 
-	int scale = 1;
+	// if we know what output we are rending on, update our
+	// cached version of the scale
 	if (surface->surface_output != NULL) {
-		scale = surface->surface_output->scale;
+		surface->scale = surface->surface_output->scale;
 	}
 
-	surface->current_buffer =
-		get_next_buffer(state->shm, surface->buffers,
-		surface->width * scale, surface->height * scale);
+	surface->current_buffer = get_next_buffer(
+		state->shm, surface->buffers,
+		surface->width * surface->scale,
+		surface->height * surface->scale);
+
 	if (surface->current_buffer == NULL) {
 		fprintf(stderr, "no buffer available\n");
 		return;
 	}
 
 	struct mako_output *output = get_configured_output(surface);
-	int height = render(surface, surface->current_buffer, scale);
+	int height = render(surface, surface->current_buffer, surface->scale);
 
 	// There are two cases where we want to tear down the surface: zero
 	// notifications (height = 0) or moving between outputs.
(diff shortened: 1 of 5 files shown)


emersion:
If this is based on #393, can you add Co-authored-by trailers where applicable to preserve authorship?

Nefsen402:
The last commit has an added Co-authored-by tag, the commit that I took direct inspiration from changes in #393.

emersion:
I'm not a fan of "wayland: Cache scale so that first frame respects HiDPI", because it doesn't work in the general case. I'd prefer to reach a consensus on https://gitlab.freedesktop.org/wayland/wayland/-/issues/133 and fix the issue in all cases instead.

wayland.c
```diff
@@ -376,11 +376,16 @@ static void layer_surface_handle_configure(void *data,
 		uint32_t serial, uint32_t width, uint32_t height) {
 	struct mako_surface *msurface = data;
 
+	zwlr_layer_surface_v1_ack_configure(surface, serial);
```
emersion, line 379: `ack_configure` must always be followed by `commit`. If the old buffer is suitable, we can perform a commit without attaching a new buffer.

emersion:
"mako: Convert signed to unsigned types for sizes" → we need to be a bit careful about this, because surprising things happen with unsigned types. Add/substract to a signed integer for instance and you can hit overflows.

wayland.c
Nefsen402: I tried to add a wl_surface_commit.

emersion: Yup, sounds good!

Nefsen402:
> "mako: Convert signed to unsigned types for sizes" → we need to be a bit careful about this, because surprising things happen with unsigned types. Add/substract to a signed integer for instance and you can hit overflows.

I needed to make these changes because otherwise the same-size commit commit would not compile as unsigned numbers cannot be compared to signed ones. I thought casting wasn't a good option, so I changed the types.

> I'm not a fan of "wayland: Cache scale so that first frame respects HiDPI", because it doesn't work in the general case. I'd prefer to reach a consensus on https://gitlab.freedesktop.org/wayland/wayland/-/issues/133 and fix the issue in all cases instead.

It's definitely not ideal, but this commit was able to get me up and running and removing the jank with current mako. Maybe we can put the caching into its own PR so that the same-size commit changes can make their way through? The other changes improve  https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3352.

Nefsen402:
> Yeah, I'd like to merge the same-size commit changes.

I pulled that out into its own PR, so that source branch names would make some sense.

emersion:
Closing because this is better addressed by listening to the `preferred_buffer_scale` event and letting the compositor dictate the scale, rather than relying on Sway's specific implementation of enter events.

PR #410, opened 2022-01-20
wayland: Don't render a frame on same-size configure

Nefsen402:
Co-authored-by: lilydjwg <lilydjwg@gmail.com>

See: https://github.com/emersion/mako/pull/409. Based on a patch in: #393

### DIFF
--- a/wayland.c
+++ b/wayland.c
@@ -376,11 +376,19 @@ static void layer_surface_handle_configure(void *data,
 		uint32_t serial, uint32_t width, uint32_t height) {
 	struct mako_surface *msurface = data;
 
+	zwlr_layer_surface_v1_ack_configure(surface, serial);
+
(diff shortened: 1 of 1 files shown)


wayland.c
```diff
@@ -376,11 +376,19 @@ static void layer_surface_handle_configure(void *data,
 		uint32_t serial, uint32_t width, uint32_t height) {
 	struct mako_surface *msurface = data;
 
+	zwlr_layer_surface_v1_ack_configure(surface, serial);
+
+	if (msurface->configured &&
+		msurface->width == (int32_t) width &&
```
emersion, line 382: Style: use two indent levels for `if` continuation lines, to tell apart the condition and the block

emersion (approved):
Thanks!

Issue #282, opened 2020-06-24
Whitespaces in config cause errors


roelofr:
I've been struggling with this issue, until I looked into the `config.c` file, but it seems that trailing whitespace breaks some config values (like colors).

## Expectations

No error is thrown with this config

```conf
background-color=#FF0000AA 
```

(there's a space after the hex)

## Result

> `Failed to parse option 'background-color'`

Likely has to do with the string length check here:
https://github.com/emersion/mako/blob/3aba143ff3f3c2acdabf86a3b6755dc55bf7248d/types.c#L45-L48

Maybe the values read from the config should be trimmed?

roelofr:
Of course, removing trailing whitespace is good practice as a user, but it shouldn't cause errors if it didn't happen.

emersion:
See https://github.com/emersion/mako/issues/277

roelofr:
I did see that issue, but figured that'd be about ` = ` vs `=`. Different issue, same result.

TheAvidDev:
The resolution depends on what we want to apply this to. Currently, `format` and (maybe) `icon-path` can have trailing whitespace. At least for `format`, I don't think there's really any effect on rendering. I'm not sure of the effect on `icon-path`. If we are okay with stripping whitespace on *both*, then this is probably an alright suggestion considering it could be applied to all options. 

However, if we allow whitespace on `format` and `icon-path` but strip it from all other options, the code would become more complicated and I'm not sure if this would be a great idea.

th0rn:
A simple solution would be to just change the error to:

Failed to parse option 'background-color' value '#FF0000AA '

That way it would be obvious to also those that don't have trailing space highlighting in their editors.

TheAvidDev:
> I'm not sure of the effect on `icon-path`. 

Just thinking back on this, issues would only arise if your icon path had a space on the end, in which case.... why does your file name contain a space with nothing after it?

Therefore, applying this stripping to all options seems alright.

emersion:
Fixed in https://github.com/emersion/mako/pull/411

PR #411, opened 2022-01-23
config: ignore leading and trailing whitespace

### LINKED ISSUE #107: strip `'` and `"` from config
note sure if a bug, but took some trial and error to figure it out what was wrong:

when converting from command line to the config file i noticed that the parser fails if e.g. the color is surrounded by quotes. 
`--border-color='#03E7B055'` -> `border-color='#03E7B055'` 

(also white spaces at the end of the line cause a fail)

pranjalkole:
Leading and trailing whitespace is now stripped before parsing the line.

getline returns line length, so strlen calls have been replaced with
that. Also, getline now runs until -1 is returned.

Additionally, the entire line (whitespace stripped) is now
printed to stderr when mako fails to parse an option.

Closes #107 and #282

### DIFF
--- a/config.c
+++ b/config.c
@@ -778,17 +778,38 @@ int load_config_file(struct mako_config *config, char *path) {
 		wl_container_of(config->criteria.next, criteria, link);
 
 	size_t n = 0;
-	while (getline(&line, &n, f) > 0) {
+	ssize_t len = 0;
+	while ((len = getline(&line, &n, f)) != -1) {
 		++lineno;
-		if (line[0] == '\0' || line[0] == '\n' || line[0] == '#') {
+
+		// We can't operate on `line` because we need to free it, so
+		// make a copy
+		char *stripped_line = line;
+
+		// Strip leading whitespace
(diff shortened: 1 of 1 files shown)


pranjalkole:
Oops, there is a bug in this. If there is no newline after the last line, the last character is not read (since len is decremented by 1). I'm rebasing and force-pushing the fix.

pranjalkole:
I thought I would implement the "strip leading whitespace" in another commit, but it was small, so added here.

PS: Sorry for the noise

murlakatamenka:
Yeah, guess C is hard if stripping whitespaces from left/right/both ends of the string requires copying of the whole string and all the efforts.

In the meantime, users are stuck with the config that is easier to parse for machines rather than for them to read.  ` = ` (`=` wrapped with whitespaces) is used for assignments in all the C sources :disappointed_relieved:

/* rant off, sorry */

pranjalkole:
ping @emersion

emersion (approved):
Thanks!

PR #419, opened 2022-03-27
contrib/completions: add bash completions

ersen0:
I didn't touch build files since I am not familiar with meson. Thanks.


emersion:
You can look at Sway for a Meson integration example.

ersen0:
@emersion thanks for pointing out, I am not sure if I did it right, I copied from Sway's meson.build file and changed it by looking at fish options here.

emersion (approved):
Thanks!

PR #421, opened 2022-04-14
Store multiple modes

towoe:
I added this to mostly allow different themes (dark/light) and the ability to add a do not disturb mode.
This allows me to add a mode `+silent` and delete `-silent` it again without having to know the current state of the theme.
But I guess it could be also used in different ways, I did not think about all use cases and all the problems which could arise here.
Just wanted to share this and hear if that would be something that would make sense to or [not](https://github.com/emersion/mako/issues/335#issuecomment-831437588).

No extra documentation for now, a little bit in the commit message, just wanted to get some feedback for now if it makes sense to go in this direction, thanks.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -58,6 +58,18 @@ static bool match_regex_criteria(regex_t *pattern, char *value) {
 	return true;
 }
 
+bool check_in_list(const char *key, char **const list) {
+	char **lp = list;
+	while (*lp) {
+		if (strcmp(key, *lp) == 0) {
+			fprintf(stderr, "check_in_list: found matching mode %s\n", *lp);
+			return true;
+		}
+		lp++;
+	}
+	return false;
+}
+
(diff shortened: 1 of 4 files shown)


emersion:
In principle this looks like a good idea to me.

- Maybe use `wl_array` instead of hand-rolling our own logic.
- I'm not a fan of handling the makoctl details (`+`/`-`) inside mako itself.
- `-` can be confused with a CLI option.
- How does this interact with https://github.com/emersion/mako/pull/382?
- Should we care about existing user-defined modes starting with `+`/`-` which this PR will break? I'd bet nobody has done that yet, so probably not.

emersion:
Alternative implementation: https://github.com/emersion/mako/pull/433

emersion:
Thanks for your PR, but #433 has superseded this patch.

towoe:
Thanks! And sorry for not updating this, currently travelling...

PR #426, opened 2022-05-05
contrib/apparmor: drop AppArmor profile

emersion:
We get tons of bug reports about it. It causes more harm than good.


Emantor (approved):

Emantor:
Makes sense to me. Those profiles can be distribution specific and unless we find someone who is willing to provide community support we should drop this.

PR #433, opened 2022-06-27
Add support for multiple modes

emersion:
Supersedes #382 and #421

I tried to split the changes in separate commits to (hopefully) make the review a bit easier.

The `makoctl` implementation really stretches POSIX sh to its limits. We should consider rewriting it in C.

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -22,6 +22,7 @@ usage() {
 	echo "  list                           List notifications"
 	echo "  reload                         Reload the configuration file"
 	echo "  set-mode <name>                Switch the current mode"
+	echo "  mode                           List modes"
 	echo "  help                           Show this help"
 }
 
@@ -37,6 +38,13 @@ fi
 call() {
 	$BUSCTL -j --user call org.freedesktop.Notifications /fr/emersion/Mako \
 		fr.emersion.Mako -- "$@"
+}
+
+require_jq() {
+	if ! type jq >/dev/null 2>&1; then
+		echo >&2 "$0: jq is required to use this command"
+		exit 1
+	fi
 }
 
 if [ $# -eq 0 ]; then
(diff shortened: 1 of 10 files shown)


makoctl
```diff
@@ -130,9 +136,50 @@ case "$1" in
...
+		shift $(($OPTIND - 1))
+		modes="$(jq -n '$ARGS.positional' --args "$@")"
+	fi
+	if [ "$add_remove_flag" = 1 ] || [ "$set_flag" = 1 ]; then
+		modes="$(echo "$modes" | jq '. | unique')"
+		modes_len="$(echo "$modes" | jq '. | length')"
+		modes_args=$(echo "$modes" | jq -r '@sh')
+		# Required to behave properly when mode names contain spaces
+		eval set -- $modes_args
+		call SetModes "as" "$modes_len" "$@"
```
vilhalmer, line 179: If the last mode is removed, should we automatically set the modes list to just `default`?

vilhalmer (changes requested):
I am admittedly pretty rusty at this point, but this looks really clean. Just one comment about the default mode.

makoctl
vilhalmer: The only downside of getting rid of it that I can think of is that there would be no way to write a config block that takes effect when there are zero modes active, since `mode=` would try to match a mode named empty string. I don't know if there is actually a usecase for this, but perhaps it would be more useful now that multiple modes are supported than it would have been before?

We could of course also special-case `mode=` to select for the empty list.

emersion: I don't see a use-case for this TBH.

emersion: Done. BTW, if someone comes up with a good use-case for a default mode, we could add something like `mako --initial-mode default` to set the initial list of modes.

vilhalmer (approved):

emersion:
Thanks for having a look!

PR #434, opened 2022-06-28
Move default config to installed file

emersion:
- [ ] This doesn't work well during development: `/etc/mako/config` is used as defaults, the user-provided config/flags are loaded on top, but `/etc/mako/config` might not be installed in which case mako won't run at all.
- [ ] Write docs

### DIFF
--- a/config.c
+++ b/config.c
@@ -722,21 +722,10 @@ static char *get_config_path(void) {
 	return NULL;
 }
 
-int load_config_file(struct mako_config *config, char *config_arg) {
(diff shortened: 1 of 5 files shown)


RaoulHC:
Would it make sense to add `SYSCONFDIR "mako/config"` as an option [here](https://github.com/emersion/mako/blob/master/config.c#L696) and provide a default file that can go there, but leave the rest of the default setting as is? That way it will still work even if it isn't present (and it probably shouldn't assume that it will be).

PR #438, opened 2022-07-29
Add app-name pattern matching

Nairou:
Needed for apps (like Telegram) that add a notification count to their window title, making exact matching fail.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -32,6 +32,7 @@ void destroy_criteria(struct mako_criteria *criteria) {
 
 	finish_style(&criteria->style);
 	free(criteria->app_name);
+	regfree(&criteria->app_name_pattern);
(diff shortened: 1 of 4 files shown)


emersion:
Does Telegram set the desktop-entry field? This would be more appropriate to check if set.

emersion:
Hm, it sounds like it's missing from the `makoctl list` output at the moment. I've added it in https://github.com/emersion/mako/pull/439.

emersion:
Is this still something you need? Or is desktop-entry working for you?

Nairou:
Didn't realize this PR was still open. There is no `desktop-entry` that I've been able to make use of, but a combination of `app-id` and `summary` have worked well enough for my needs. Thanks!

Issue #441, opened 2022-08-24
Segmentation fault after updating wlroots/sway


kamiyaa:
mako git version 8417d1400b4ab3dd496da2cbfec7b21041833a7b

wlroots git version [d3cb6b73a2733d8457a8be2593a42942b36b77e2](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/d3cb6b73a2733d8457a8be2593a42942b36b77e2)
sway git version [6e023257e8e9a6673fc3ff69048cdaf8baa70044](https://github.com/swaywm/sway/commit/6e023257e8e9a6673fc3ff69048cdaf8baa70044)

I'm not sure what the exact cause is because the segfaults are unpredictable; it just happens a few minutes after starting `mako` and receiving some notifications.
I modified mako to get a rough idea of where the segfault is happening and it seems to be segfaulting at `wl_display_dispatch`:

`event-loop.c`
```c
...

if (loop->fds[MAKO_EVENT_WAYLAND].revents & POLLIN) {
      printf("wl_display_dispatch\n");
      printf("loop %p\n", loop);
      printf("loop->display %p\n", loop->display);
      ret = wl_display_dispatch(loop->display);
      if (ret < 0) {
            fprintf(stderr, "failed to read Wayland events\n");
            break;
      }
}
printf("Done\n");

...
```

### Output
```
...
Done
loop
Done
loop
wl_display_dispatch
loop 0x7ffe83573840
loop->display 0x562debf12bd0
Done
loop
wl_display_dispatch
loop 0x7ffe83573840
loop->display 0x562debf12bd0
Segmentation fault
```

emersion:
Please obtain a stack trace.

kamiyaa:
@emersion how do I obtain a stacktrace? I'm not very familiar with C tooling

kamiyaa:
```bash
Thread 1 "mako" received signal SIGSEGV, Segmentation fault.
pointer_handle_enter (data=0x555555590f90, wl_pointer=0x555555591d00, serial=1950, wl_surface=0x5555555ade50, surface_x=136966, surface_y=52603) at ../wayland.c:232
232			if (surface->surface_output->scale > scale) {
(gdb) backtrace
#0  pointer_handle_enter
    (data=0x555555590f90, wl_pointer=0x555555591d00, serial=1950, wl_surface=0x5555555ade50, surface_x=136966, surface_y=52603) at ../wayland.c:232
[...]

Issue #440, opened 2022-08-11
SIGSEGV when hovering over a notification under hyprland


pinkcreeper100:
Unsure where coredumps are stored

emersion:
Please post a stack trace. "It segfaults" isn't very useful info.

pinkcreeper100:
where do I find a stack trace?

pinkcreeper100:
```
Aug 11 19:48:40 pc100 systemd-coredump[2090]: Process 2072 (mako) of user 1000 dumped core.
                                              Module mako with build-id 91c64eb80b15c4cdbd2180e7079624efc322223b
                                              #0  0x00005595570505e8 n/a (mako + 0xa5e8)
                                              #6  0x000055955704d8b0 n/a (mako + 0x78b0)
                                              #9  0x000055955704dc65 n/a (mako + 0x7c65)
```

pinkcreeper100:
```
           PID: 2072 (mako)
           UID: 1000 (pinkcreeper100)
           GID: 1000 (pinkcreeper100)
        Signal: 11 (SEGV)
     Timestamp: Thu 2022-08-11 19:48:40 BST (10min ago)
  Command Line: mako
    Executable: /usr/bin/mako
 Control Group: /user.slice/user-1000.slice/session-1.scope
          Unit: session-1.scope
         Slice: user-1000.slice
       Session: 1
     Owner UID: 1000 (pinkcreeper100)
       Boot ID: 09c37dc2ca874d71b813138bb621c204
    Machine ID: c520689af4a542e98a19d57f19976f1f
      Hostname: pc100
       Storage: /var/lib/systemd/coredump/core.mako.1000.09c37dc2ca874d71b813138bb621c204.2072.1660243720000000>
     Disk Size: 358.2K
       Message: Process 2072 (mako) of user 1000 dumped core.
                
                Module linux-vdso.so.1 with build-id 1ae89d98798409ec8d014bb68399cba9c70ec3b4
                Module libblkid.so.1 with build-id fb2c5d3c17aac74758a3eb80a2bc1c16bcf183b1
                Module libbrotlicommon.so.1 with build-id acfd597a977c8087bb6184383daae2e828a9ce42
                Module libgraphite2.so.3 with build-id ce58945ebb55b86d3a4e717b6eae29efc4720d8e
                Module libdatrie.so.1 with build-id 6fe3b6ece2c8e7d11869fa051375128d8f808f58
[...]

PR #442, opened 2022-08-25
fix segfault when surface->surface_output is NULL

kamiyaa:
Just adds a check to make sure `surface_output` is not NULL

I believe this fixes https://github.com/emersion/mako/issues/441 and https://github.com/emersion/mako/issues/440

Feedback is appreciated

### DIFF
--- a/wayland.c
+++ b/wayland.c
@@ -229,6 +229,9 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 
 	struct mako_surface *surface;
 	wl_list_for_each(surface, &state->surfaces, link) {
+		if (!surface->surface_output) {
+			continue;
+		}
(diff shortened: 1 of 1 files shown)


wayland.c
```diff
@@ -229,6 +229,9 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 
 	struct mako_surface *surface;
 	wl_list_for_each(surface, &state->surfaces, link) {
+		if (!surface || !surface->surface_output) {
```
emersion, line 232: The `surface` NULL check is unnecessary.

emersion (commented):
LGTM apart from this comment!

wayland.c
kamiyaa: Fixed!

emersion (approved):
Thanks for the fix!

PR #444, opened 2022-09-14
makoctl: add command to list history

Gusser93:
Implements `makoctl history` discussed in #91.
Extends #220 by listing the history buffer with `makoctl history`in the same format as `makoctl list`. I put the code from `dbus/mako.c:handle_list_notifications` into  `dbus/mako.c:handle_list` and made the list given to `wl_list_for_each` an argument. `handle_list`is than used by `handle_list_notification` with `mako_state->notifications` and by `handle_list_history` with `mako_state->history`. A short test was successful, but I should note that I have no experience with dbus.

### DIFF
--- a/contrib/completions/bash/makoctl
+++ b/contrib/completions/bash/makoctl
@@ -11,6 +11,7 @@ _makoctl()
     'invoke'
     'menu'
     'list'
+    'history'
     'reload'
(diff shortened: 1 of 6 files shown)


emersion (approved):
Looks good to me!

PR #445, opened 2022-10-19
Fix positioning of icon and text

### LINKED ISSUE #329: Bottom padding is shorter than expected (regression)
Hey team, just upgraded to commit 6d743d272e277c74246642f14b5ba2a8fc202919 and am having the following regression.

I have `padding=16,32` in my config. Here's the normal, expected padding from version 1.4.1:

good padding

Here's the problem from the commit above, where the bottom padding seems to be shorter than normal:

bad padding

PlayerNameHere:
This PR fixes two things:
- The positioning of the icon &mdash; see the commit message for more information
- The positioning of the text, which was offset by the size of the border, either along the x-asis or y-axis depending on whether the icon is positioned top/bottom or left/right. This should fix #329.

### DIFF
--- a/render.c
+++ b/render.c
@@ -294,9 +294,9 @@ static int render_notification(cairo_t *cairo, struct mako_state *state, struct 
 	}
 
 	if (icon_vertical) {
-		text_x = (notif_width - text_width) / 2;
+		text_x = (notif_width - text_width - border_size) / 2;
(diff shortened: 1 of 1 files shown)


loserMcloser:
Please merge this!

emersion (approved):
Thanks!

PR #447, opened 2022-10-21
Add urgency indicator in makoctl list.

### LINKED ISSUE #446: FR: Provide urgency in makoctl list output
I would like to make a script that gives me the count of current notifications, but ignoring low urgency ones.

I can get the total count with something like `makoctl list | jq '.data[] | length'`, however the output does not specify the urgency so I can filter them out.

Would it be possible to also have an entry like this inside the payload for each notification?

```json
"urgency": {
  "type": "s",
  "data": "critical"
}
```

Or, if not, is there any other method I could use to achieve my use case?

Jadarma:
Adds a new `urgency` field for each notification in the `makoctl list` output.

The value returned is `0` for `Low`, `1` for `Normal`, and `2` for `Critical`, same as the spec.

An example output for `notify-send -u critical 'The house is on fire!'`:
```json
{
        "type" : "aa{sv}",
        "data" : [
                [
                        {
                                "app-name" : {
                                        "type" : "s",
                                        "data" : "notify-send"
                                },
                                "app-icon" : {
                                        "type" : "s",
                                        "data" : ""
                                },
                                "category" : {
                                        "type" : "s",
                                        "data" : ""
                                },
                                "desktop-entry" : {
                                        "type" : "s",
                                        "data" : ""
                                },
                                "summary" : {
                                        "type" : "s",
                                        "data" : "The house is on fire!"
                                },
                                "body" : {
                                        "type" : "s",
                                        "data" : ""
                                },
                                "id" : {
                                        "type" : "u",
                                        "data" : 1
                                },
                                "urgency" : {
                                        "type" : "y",
                                        "data" : 2
                                },
                                "actions" : {
                                        "type" : "a{ss}",
                                        "data" : {}
                                }
                        }
                ]
        ]
}
```

Closes #446

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -202,6 +202,12 @@ static int handle_list(sd_bus_message *msg, struct wl_list *list) {
 			return ret;
 		}
 
+		ret = sd_bus_message_append(reply, "{sv}", "urgency",
+			"y", notif->urgency);
+		if (ret < 0) {
+			return ret;
+		}
+
 		ret = sd_bus_message_open_container(reply, 'e', "sv");
 		if (ret < 0) {
 			return ret;


emersion (approved):
LGTM, thanks!

PR #453, opened 2023-02-06
Add `invoke-action <action>`

Consolatis:
Before this patch, it was impossible to send a different action
than `default` when clicking on a notification without loosing
access to the surface and thus the xdg_activation token.

With this patch, the user is now able to configure the action
to be sent when clicking on the notification while preserving
access to the xdg_activation token.

Example configuration:
```ini
[app-name="some-app-id" actionable]
on-button-left=invoke-action mail-reply-sender

### DIFF
--- a/config.c
+++ b/config.c
@@ -126,7 +126,8 @@ void init_default_style(struct mako_style *style) {
 	style->anchor =
 		ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP | ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
 
-	style->button_bindings.left.action = MAKO_BINDING_INVOKE_DEFAULT_ACTION;
+	style->button_bindings.left.action = MAKO_BINDING_INVOKE_ACTION;
+	style->button_bindings.left.action_name = strdup(DEFAULT_ACTION_KEY);
 	style->button_bindings.right.action = MAKO_BINDING_DISMISS;
 	style->button_bindings.middle.action = MAKO_BINDING_NONE;
 	style->touch_binding.action = MAKO_BINDING_DISMISS;
@@ -141,6 +142,7 @@ void init_empty_style(struct mako_style *style) {
 
 static void finish_binding(struct mako_binding *binding) {
 	free(binding->command);
+	free(binding->action_name);
 }
 
 void finish_style(struct mako_style *style) {
@@ -162,6 +164,9 @@ static void copy_binding(struct mako_binding *dst,
 	*dst = *src;
 	if (src->command != NULL) {
 		dst->command = strdup(src->command);
+	}
+	if (src->action_name != NULL) {
+		dst->action_name = strdup(src->action_name);
 	}
 }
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: include/config.h [hunk 1]
```diff
@@ -12,13 +12,17 @@ enum mako_binding_action {
...
 	MAKO_BINDING_EXEC,
 };
 
 struct mako_binding {
 	enum mako_binding_action action;
-	char *command; // for MAKO_BINDING_EXEC
+	union {
+		char *command;     // for MAKO_BINDING_EXEC
+		char *action_name; // for MAKO_BINDING_ACTION
+	};
```
Consolatis, line 25: This is up for discussion, its saves a pointer but makes `free(binding->command)` a bit weird when used for `MAKO_BINDING_INVOKE_ACTION`.

Alternatives:
- rename to something generic like `data`
- just add a 2nd pointer
- reuse the `command` pointer

config.c
```diff
@@ -647,6 +647,9 @@ static bool apply_style_option(struct mako_style *style, const char *name,
 			binding.action = MAKO_BINDING_DISMISS_GROUP;
 		} else if (strcmp(value, "invoke-default-action") == 0) {
 			binding.action = MAKO_BINDING_INVOKE_DEFAULT_ACTION;
+		} else if (has_prefix(value, "invoke-action ")) {
+			binding.action = MAKO_BINDING_INVOKE_ACTION;
+			binding.action_name = strdup(value + strlen("invoke-action "));
```
Consolatis, line 652: Instead of using a separate action we could also implement `invoke-default-action` via the new `invoke-action` binding.
It should just require moving `DEFAULT_ACTION_KEY` over to `include/config.h`.

emersion:
Looks pretty good to me!

include/config.h
emersion: I'd prefer to just use two separate fields without `union`.

config.c
emersion: Yeah, that sounds like a good idea!

Consolatis: Hm, I am getting a double free on `finish_binding()`, likely because of the initial `left` binding now containing something else than `NULL`. Will keep debugging this later.

include/config.h
Consolatis: changed

notification.c
```diff
@@ -364,21 +383,9 @@ void notification_execute_binding(struct mako_notification *notif,
...
-				}
-				notify_action_invoked(action, activation_token);
-				free(activation_token);
-				break;
-			}
-		}
-		close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
+	case MAKO_BINDING_INVOKE_ACTION:
+		assert(binding->action_name != NULL);
+		try_invoke_action(notif, binding->action_name, ctx);
```
Consolatis, line 388: As there is only a single user now, should this keep being a function call or should I move `try_invoke_action()` back into the switch?

config.c
Consolatis: addressed

notification.c
emersion: I think having a separate function is worthwhile, because of the indentation level.

emersion (approved):
LGTM!

Consolatis:
I think all issues were addressed, Squashed + fixed a wrong comment.

emersion:
Thanks!

PR #457, opened 2023-03-14
chore: check iconname before pass it to pattern

Decodetalkers:
resolve: https://github.com/emersion/mako/issues/456

https://github.com/emersion/mako/blob/master/dbus/xdg.c#L170-L178

when it happened, the ret is  always 0, then I think it may not be the problem of dbus handle. and the message is send cannot be control, then I think maybe should check the icon_name,before, for the badname

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -20,6 +20,27 @@
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include "cairo-pixbuf.h"
 
+static bool validate_icon_name(const char* icon_name) {
+	int icon_len = strlen(icon_name);
+	if (icon_len > 1024) {
+		return false;
+	}
+	int index;
+	for (index = 0; index < icon_len; index ++) {
+		bool is_number = icon_name[index] >= '0' && icon_name[index] <= '9';
+		bool is_abc = (icon_name[index] >= 'A' && icon_name[index] <= 'Z') ||
+				(icon_name[index] >= 'a' && icon_name[index] <= 'z');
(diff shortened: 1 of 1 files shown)


Decodetalkers:
or maybe it is just the problem of glob.?

emersion:
It sounds like the lowercase restriction is a bit too much, there are some icons with upper-case letters, e.g. `org.gnome.Calculator`. The rest sounds fine.

Decodetalkers:
> From https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html:
> 
> > all icon names may only contain lowercase letters, numbers, underscore, dash, or period characters
> 
> Maybe we can check this instead?

I use regex now

emersion:
I'd prefer not to use a regexp, and instead have a simple loop to validate each character in the string.

Can we also invert the function's return value and rename it to e.g. `validate_icon_name()`?

icon.c
```diff
@@ -20,6 +21,15 @@
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include "cairo-pixbuf.h"
 
+static bool bad_iconname(const char* icon_name) {
+	if (strlen(icon_name) > 30) {
```
emersion, line 25: I'd prefer to pick a much higher limit here, e.g. 1024.

Decodetalkers: ok, do as your suggestion

emersion (approved):
Thanks!

PR #458, opened 2023-03-16
mako.5.scd: enhance the use of mpv

melentan:
Add the mpv --keep-open=no option to prevent too many connections to the pulseaudio audio server. That case occurs where user has keep-open=yes in their mpv configuration. Each mpv occurrence will stay logged to the pulseaudio server, and prevent a new connection once 64 clients connected.

The --keep-open=no option allows to close the mpv option once the sound played.

### DIFF
--- a/mako.5.scd
+++ b/mako.5.scd
@@ -61,7 +61,7 @@ on-button-middle=exec makoctl menu -n "$id" dmenu -p 'Select action: '
 The following option will play a sound when a new notification is opened:
 
 ```
-on-notify=exec mpv /usr/share/sounds/freedesktop/stereo/message.oga
+on-notify=exec mpv --keep-open=no /usr/share/sounds/freedesktop/stereo/message.oga
 ```
 
 When using _invoke-action_ or _invoke-default-action_ a xdg-activation


emersion:
I believe `--keep-open=no` is the default?

melentan:
> I believe `--keep-open=no` is the default?

Yes: https://github.com/mpv-player/mpv/blob/master/etc/builtin.conf#L35 (file embedded into the mpv binary at compile time)

emersion:
Oh, sorry, I skipped the PR description for some reason.

I'd rather not reset all options to their defaults just in case the user has changed them. The same argument could be made for a lot of other options.

melentan:
I think the idea is not to reset the `--keep-open` option to its default value, but simply to set it in order to prevent having too many `pulseaudio` clients in case where user has `--keep-open=yes`. The user should not care about a mpv instance run by `mako` just to play a brief notification sound.
After, maybe just a comment inside the configuration file notifying the user about this case?

emersion:
Sorry, but this dwells too much into user-specific configuration issues for my taste. Thanks for sending a PR though!

PR #460, opened 2023-04-08
use last_scale initially to avoid many redraws (which cause flickering)

lilydjwg:
## The problem

mako appears with scale=1, and finds the output prefers scale=2, so it redraws with scale=2. This makes the window reshapes (supposingly because pango redraws with higher dpi and the rounding is different). So every time it appears, it flickers once.

## Proposed solution

Remember last scale because it's very likely that the next time a notification is shown it's on the same output, thus avoiding above flickering in this case.

### DIFF
--- a/include/mako.h
+++ b/include/mako.h
@@ -75,6 +75,7 @@ struct mako_state {
 	struct wl_list notifications; // mako_notification::link
 	struct wl_list history; // mako_notification::link
 	struct wl_array current_modes; // char *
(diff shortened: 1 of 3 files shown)


emersion:
The real solution is adding support for the new `wl_surface.preferred_buffer_scale` event. I'd rather not add fragile heuristics.

lilydjwg:
OK, I'll wait for it to settle.

PR #461, opened 2023-04-12
Use systemd to check environment variable

akay:
Unit files have a built-in condition to check variables, so let's use it instead of executing a shell command.


akay:
Seems like there are some issues with this approach that I'm not experiencing, so I will dig into it.

PR #471, opened 2023-07-29
Generalize String criteria matching

### LINKED ISSUE #306: Generalize criteria regex matching
Currently `summary` and `body` can use the `~=` operator for a POSIX regex match.

Generalize this so that any string criteria can be matched using a regex. Could replace `strcmp` calls with a function that accounts for both.

earboxer:
Closes #306

Most string fields have been changed to support

* Regex matching `[app-name~=D..o]`
* Not equal to `[summary!=""]`
* Bare specification `[summary]` (equivalent to `[summary!=""]`)
* Bare negation `[!summary]` (equivalent to `[summary=""]`)

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -27,18 +27,34 @@ struct mako_criteria *create_criteria(struct mako_config *config) {
 	return criteria;
 }
 
+void free_cond(struct mako_condition *cond) {
+	switch(cond->operator) {
+		case OP_EQUALS:
+		case OP_NOT_EQUALS:
+			free(cond->value);
+			return;
+		case OP_REGEX_MATCHES:
+			regfree(&cond->pattern);
+			return;
+		case OP_NONE:
+		case OP_TRUTHY:
+		case OP_FALSEY:
+		default:
+			// Nothing to free.
+			return;
+	}
+}
+
 void destroy_criteria(struct mako_criteria *criteria) {
 	wl_list_remove(&criteria->link);
 
 	finish_style(&criteria->style);
-	free(criteria->app_name);
-	free(criteria->app_icon);
-	free(criteria->category);
-	free(criteria->desktop_entry);
-	free(criteria->summary);
-	regfree(&criteria->summary_pattern);
-	free(criteria->body);
-	regfree(&criteria->body_pattern);
+	free_cond(&criteria->app_name);
+	free_cond(&criteria->app_icon);
+	free_cond(&criteria->category);
+	free_cond(&criteria->desktop_entry);
+	free_cond(&criteria->summary);
+	free_cond(&criteria->body);
 	free(criteria->raw_string);
 	free(criteria->output);
 	free(criteria->mode);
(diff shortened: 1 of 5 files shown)


earboxer:
I included the operator `OP_NONE`, but it's not being used.

Originally I was thinking this could replace the 'spec' definitions, but didn't dig deep enough to see if this is the case or not...

include/criteria.h
```diff
@@ -11,6 +11,14 @@
 struct mako_config;
 struct mako_notification;
 
+enum operator { OP_NONE, OP_EQUALS, OP_REGEX_MATCHES, OP_NOT_EQUALS, OP_TRUTHY, OP_FALSEY };
+
+struct mako_condition {
```
emersion, line 16: This is specific to strings, maybe should be renamed to e.g. `struct mako_string_criteria`?

include/criteria.h
```diff
@@ -11,6 +11,14 @@
...
+	enum operator operator;
+	char* value;
```
emersion, line 18: Style nit: `*` should be on the right

criteria.c
```diff
@@ -27,18 +27,34 @@ struct mako_criteria *create_criteria(struct mako_config *config) {
 	return criteria;
 }
 
+void free_cond(struct mako_condition *cond) {
+	switch(cond->operator) {
```
emersion, line 31: Style nit: missing space between `switch` and `(`

criteria.c
```diff
@@ -27,18 +27,34 @@ struct mako_criteria *create_criteria(struct mako_config *config) {
...
+		case OP_EQUALS:
```
emersion, line 32: Style nit: the whole `switch` block should have one less indent level.

criteria.c
```diff
@@ -27,18 +27,34 @@ struct mako_criteria *create_criteria(struct mako_config *config) {
...
+		case OP_NOT_EQUALS:
+			free(cond->value);
+			return;
+		case OP_REGEX_MATCHES:
+			regfree(&cond->pattern);
+			return;
+		case OP_NONE:
+		case OP_TRUTHY:
+		case OP_FALSEY:
+		default:
```
emersion, line 42: Could we drop this `default`? That way, the compiler will warn when we're missing a case.

criteria.c
```diff
@@ -59,6 +75,24 @@ static bool match_regex_criteria(regex_t *pattern, char *value) {
...
+		case OP_REGEX_MATCHES:
+			return match_regex_criteria(&cond->pattern, value);
+		case OP_TRUTHY:
+			return strcmp("", value) != 0;
+		case OP_FALSEY:
+			return strcmp("", value) == 0;
+		case OP_NONE:
+			return true;
+	}
+	return true;
```
emersion, line 93: Probably should be an `abort()`, since this is unreachable?

emersion (commented):
Thanks for the patch! Sorry I took way too much time to get back to this.

I think it's a bit confusing to introduce all of these new operators, but only support them on strings. For instance truthy/falsy are not supported on booleans.

Maybe operators could be introduced incrementally on all types they make sense, starting with e.g. regexp which only makes sense on strings.

include/criteria.h
earboxer: I think I was originally thinking that mako_condition could be used for all of the conditions (and maybe there are other operators that other types of conditions could use, like "in" or "less than"/"greater than" for urgency, group-index).

Since this PR is "Generalize String criteria matching", I didn't want to go too far out of scope and refactor all of the conditions.

But I'm not too invested in this, so if you really want the rename, I will probably do it.

emersion: How would that work? The `value` field is a `char *`, but we'd need some other type for urgency/group-index/etc.

PR #477, opened 2023-10-02
Allow forgetting notifications

### LINKED ISSUE #164: Add a way to ignore notifications
Add a way to drop notifications matching some criteria.

I think @AN3223 is interested in this.

mhmdanas:
Fixes #164.

Tested briefly and works fine, but bear in mind that this is a very quick and quite possibly deficient implementation, as I'm unfamiliar with C and Mako's codebase, therefore code review and testing are very appreciated.


dbus/xdg.c
```diff
@@ -431,6 +431,10 @@ static int handle_notify(sd_bus_message *msg, void *data,
 
 	notification_execute_binding(notif, &notif->style.notify_binding, NULL);
 
+	if (notif->style.ignore) {
+		close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
```
mhmdanas, line 435: Do you think it might be a good idea to use something other than `MAKO_NOTIFICATION_CLOSE_DISMISSED` (or maybe even create an entirely new value, e.g. `MAKO_NOTIFICATION_CLOSE_IGNORED`)?

mako.5.scd
```diff
@@ -380,6 +380,12 @@ associated command-line option.
 
 	Default: 0
 
+*ignore*=0|1
+	Whether to ignore this notification. Ignored notifications are immediately
+	dismissed, and are not inserted into history.
```
mhmdanas, line 385: I can feel like this description can be improved; suggestions?

dbus/xdg.c
mhmdanas: > The app might maintain some state related to the notification (to handle actions and such), so maybe sending the Close signal is desirable. OTOH, maybe the app performs an undesirable side-effect when it receives the Close signal. (But that probably would be better reported as an app bug, rather than worked around in mako?)

Like you mention, the app doing some undesirable side effect on notification close is something to be reported upstream, so I think sending the signal is the reasonable option here.

> If we send the Close signal, on-notify=dismiss might make more sense. If we don't send the Close signal, ignore=1 might make more sense.

Yeah, but then again, `on-notify=dismiss` saves the notification to history, which we don't want (unless you meant to type `on-notify=ignore` instead?).

emersion: > Yeah, but then again, on-notify=dismiss saves the notification to history, which we don't want (unless you meant to type on-notify=ignore instead?).

Right. What I meant is that it made sense to re-use the `on-notify` option if we are to simulate a way to immediately dismiss a notification when it's opened.

Sorry to switch my preference, but in the end I like your proposal of `on-notify=delete` better I think. I'm not a huge fan of the name "delete" though, maybe "discard" would be a bit better? Or maybe `dismiss --no-history`?

mhmdanas: > Sorry to switch my preference, but in the end I like your proposal of on-notify=delete better I think. I'm not a huge fan of the name "delete" though, maybe "discard" would be a bit better? Or maybe dismiss --no-history?

`forget`?

mhmdanas: Just pushed a quick implementation, seems to be working well.

emersion: `forget` would drop an already-dismissed notification from the history I think…

mhmdanas: Done.

I'm also thinking about adding `--no-history` to `makoctl dismiss`, but since currently the option doesn't exist for `dismiss-all` and `dismiss-group`, the code will need to be modified a bit to support that.

notification.c
```diff
@@ -129,12 +130,14 @@ void close_notification(struct mako_notification *notif,
...
-		destroy_notification(n);
+	if (add_to_history) {
+		wl_list_insert(&notif->state->history, &notif->link);
+		while (wl_list_length(&notif->state->history) >
+			notif->state->config.max_history) {
+			struct mako_notification *n =
+				wl_container_of(notif->state->history.prev, n, link);
+			destroy_notification(n);
+		}
 	}
```
emersion, line 141: We need to `destroy_notification(notif)` in the else branch to avoid memory leaks.

emersion (commented):
Apart from this, looks good!

emersion (approved):
Thank you!

dbus/xdg.c
RichardFevrier: I've tried to use `--no-history` is my config without success.
Is it working on your side?

PR #478, opened 2023-10-14
render: fix segfault

pranjalkole:
The notification style was being accessed after destroying the notification struct.


emersion (approved):
Good catch, thanks!

PR #486, opened 2023-12-07
Set wayland surface width per each mako_surface

MikeWalrus:
Using the super style's width for all Wayland surfaces causes extra space to be blurred when using the compositor to blur the background of notification surfaces.

Before:
before

After:
after


MikeWalrus:
Thanks!

PR #487, opened 2023-12-11
readme: basu needs a manual dbus user session

blabber:
A manually started dbus user session is not only needed for elogind, but for basu too. In other words, any environment not using systemd might need it.

It took me a while to figure this out on my FreeBSD box.

As I am not a native speaker, feel free to correct any mistakes I made in wording or grammar.


emersion (approved):
LGTM, thanks!

PR #488, opened 2023-12-26
Remove unused includes

mhmdanas:


surface.c
```diff
@@ -2,7 +2,6 @@
 #include <string.h>
 #include <stdlib.h>
 
-#include "config.h"
```
emersion, line 5: Are we sure this won't break some `#ifdef HAVE_ICONS` for instance?

emersion: Ah, nevermind, we don't use a header for conditionals.

emersion (approved):
Thanks!

PR #490, opened 2023-12-31
Use explicit raw JSON from busctl

pnelson:
Some time ago `makoctl mode` mysteriously stopped working for me with a `jq` error that seemed to indicate bad JSON content. I finally took some time to debug it and it was because I had set `SYSTEMD_COLORS=16` for my user. The `-j` flag on `busctl` appears to be sensitive to this. Changing it to `--json=short` will ensure the output of that command is always raw.

```
$ SYSTEMD_COLORS=16 busctl -j --user call org.freedesktop.Notifications /fr/emersion/Mako fr.emersion.Mako -- ListModes | jq '.data[0]'
jq: parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 1, column 18
```

```
$ SYSTEMD_COLORS=16 busctl --json=short --user call org.freedesktop.Notifications /fr/emersion/Mako fr.emersion.Mako -- ListModes | jq '.data[0]'
[
  "default"
]
```

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -39,7 +39,7 @@ if ! type $BUSCTL >/dev/null 2>&1; then
 fi
 
 call() {
-	$BUSCTL -j --user call org.freedesktop.Notifications /fr/emersion/Mako \
+	$BUSCTL --json=short --user call org.freedesktop.Notifications /fr/emersion/Mako \
 		fr.emersion.Mako -- "$@"
 }
 


fluix-dev:
This looks good to me, and it matches the man pages:
```
--json=MODE
    When used with the call or get-property command, shows output
    formatted as JSON. Expects one of "short" (for the shortest
    possible output without any redundant whitespace or line
    breaks) or "pretty" (for a pretty version of the same, with
    indentation and line breaks). Note that transformation from
    D-Bus marshalling to JSON is done in a loss-less way, which
    means type information is embedded into the JSON object tree.

-j
    Equivalent to --json=pretty when invoked interactively from a
    terminal. Otherwise equivalent to --json=short, in particular
    when the output is piped to some other program.
```

emersion:
makoctl has been rewritten in C.

PR #492, opened 2024-01-11
Avoid invalid frame callbacks

### LINKED ISSUE #493: Showing notifications and modifying modes at the same time sometimes lead to segmentation faults.
To reproduce:
```bash
for i in {1..20}; do
    notify-send 'some notification' && makoctl mode -a some_mode
done
```

#492 fixes this.

MikeWalrus:
`surface->frame_callback` could have already been destroyed in `layer_surface_handle_closed`, and calling `wl_callback_destory` with NULL segfaults.

Also, the frame callback should be destroyed in `destroy_surface`, otherwise the surface can be used in the callback after being freed.

Fixes #493.

### DIFF
--- a/surface.c
+++ b/surface.c
@@ -12,6 +12,9 @@ void destroy_surface(struct mako_surface *surface) {
 	if (surface->surface != NULL) {
 		wl_surface_destroy(surface->surface);
 	}
+	if (surface->frame_callback != NULL) {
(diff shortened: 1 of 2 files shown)


MikeWalrus:
Fixes #493.

emersion (approved):
LGTM, thanks for the fix!

PR #494, opened 2024-01-20
Implement (hacky) support for scalable icons

Tharre:
Previously, any icons that were found in folders like scalable or symbolic were ignored because they didn't contain any sizing information. Instead, assume any icons ending with ".svg" to be scalable as it's the only supported vectorized format.

With this, the adwaita icon theme should work again, they removed most of their fixed size icons semirecently:

https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/issues/221

### DIFF
--- a/icon.c
+++ b/icon.c
@@ -170,6 +170,13 @@ static char *resolve_icon(struct mako_notification *notif) {
 		for (size_t i = 0; i < found_count; ++i) {
 			char *relative_path = icon_glob.gl_pathv[i];
 
+			if (!strcmp(relative_path+strlen(relative_path)-4, ".svg")) {
+				// Assume icon is scalable
+				free(icon_path);
+				icon_path = strdup(icon_glob.gl_pathv[i]);
+				break;
+			}
+
 			// Find the end of the current search path and walk to the next
 			// path component. Hopefully this will be the icon resolution
 			// subdirectory.


icon.c
```diff
@@ -170,6 +170,13 @@ static char *resolve_icon(struct mako_notification *notif) {
 		for (size_t i = 0; i < found_count; ++i) {
 			char *relative_path = icon_glob.gl_pathv[i];
 
+			if (!strcmp(relative_path+strlen(relative_path)-4, ".svg")) {
```
emersion, line 173: Maybe a better approach would be to check if the size subdir is named "scalable"?

Tharre: Would have to check for "scalable" and "symbolic", but yeah, possibly. There definitely seem to be themes that have more complex SVGs for bigger sizes, so not throwing away sizing information for them may be desirable.

On the other hand I don't know if other icon themes use different names for this; though spot checking the arch repos that does seem to be a common theme.

For my usecase this definitely works, if you want I can update this PR to this commit:
https://github.com/Tharre/mako/commit/f7eed4670584317078340d4a45611347afc14bb3

emersion: Hm, right, I suppose that to do things the Right Way, we'd need to parse the theme .ini files... Although, maybe we should keep parsing the size if there are .svg files in a directory with a fixed size...

emersion: We could switch to libsfdo for icon lookup.

Tharre: Sure thing. Should I whip up another PR for it or do you want to do it yourself?

emersion: Feel free to open a PR!

emersion:
Let's close this for now, I don't think I'd want to maintain logic with this approach.

PR #495, opened 2024-01-25
Extend --no-history functionality

grmat:
Hi,
the recently introduced `dismiss --no-history` feature is very helpful to me.
However, afaics, it's only usable when clicking the notification itself, not through the dbus API (or I'm blind) - and therefore neither with `makoctl`.

I tried to get familiar with the codebase to add a parameter for dbus calls and it seems a bit weird to me.
For my elaboration, I'd like to exclude `DismissAllNotifications()`, as this one is self-explanatory.
For the rest, there are currently 3 function signatures for 4 situations.
If there's an ID given, `DismissNotification()` is used to dismiss whole groups as well, not `DismissGroupNotifications()`.
I drafted a proposal of what would be easier to understand IMHO[^1]:

ID given? | Whole group? | Current                        | Proposal
---       | ---          | ---                            | ---
0         | 0            | DismissLastNotification()      | DismissLastNotification(false)
0         | 1            | DismissGroupNotifications()    | DismissLastNotification(true)
1         | 0            | DismissNotification(id, false) | DismissNotification(id, false)
1         | 1            | DismissNotification(id, true)  | DismissNotification(id, true)

[^1]: of course, with 2 params it would also be possible to handle all those cases in one function (e.g. id=0 for last), but I'm not sure if this would go too far.

I've implemented this draft in the given PR.

Features/use cases:
- choose whether _group_ and _all_ notifications should appear in history (not only single ones)
- choose whether notifications should appear in history via dbus
- activate actions not only when receiving them but also on timeout and/or dismissal
- possibility to keep an indicator showing the current num of (missed) notifications without periodically polling mako

I'm aware of the implications an API change could entail and I haven't been involved in this project. I don't claim to say that change would make sense here, I'd just like to ask

- is extending the --no-history functionality even interesting?
- is an API change even open to discussion?
- general comments?

### DIFF
--- a/config.c
+++ b/config.c
@@ -149,6 +149,7 @@ void finish_style(struct mako_style *style) {
 	finish_binding(&style->button_bindings.right);
 	finish_binding(&style->touch_binding);
 	finish_binding(&style->notify_binding);
+	finish_binding(&style->timeout_binding);
 	free(style->icon_path);
 	free(style->font);
 	free(style->format);
@@ -388,6 +389,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	if (style->spec.notify_binding) {
 		copy_binding(&target->notify_binding, &style->notify_binding);
 		target->spec.notify_binding = true;
+	}
+
+	if (style->spec.timeout_binding) {
+		copy_binding(&target->timeout_binding, &style->timeout_binding);
+		target->spec.timeout_binding = true;
 	}
 
 	return true;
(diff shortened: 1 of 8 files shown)


emersion:
Hi! Sorry for the delay.

Yeah, the current D-Bus API is pretty clunky. I wonder if we should just have a single `Dismiss` method which takes a dict of options (that way, we can extend it later more easily). Seems like this is a somewhat common D-Bus pattern?

I think it's fine to make breaking changes to our internal D-Bus API.

emersion:
Superseded by https://github.com/emersion/mako/pull/564

Issue #507, opened 2024-03-18
Improve README's Running section


o0nd7ots:
As of right now Mako targets Sway on SystemD systems and offers support for other compositors and init systems.
This support as of right now is kind of neglected (see #505  and #88) .

I will try to improve it, each commit explaining why I did something and linking to this issue.

emersion:
mako doesn't target systemd specifically. mako does assume that a D-Bus session bus is available, as required by the FreeDesktop notifications spec. systemd happens to setup a session bus by default, while in general for non-systemd distros there are extra steps to get that running. However, none of the session bus stuff is mako-specific.

o0nd7ots:
This makes sense.

I'll fix the patch as soon as I can.

-------- Original Message --------

o0nd7ots:
From what I can see this is fixed

PR #508, opened 2024-03-18
Readme: Improved support

o0nd7ots:
I have extracted "Multiple notification daemons" to another section.

Also I have added a line explaining that mako needs a session bus.

Later I have split systemd and non-systemd systems into two paragraphs

Related to #507

PS. Sorry for the fuss, I don't really have experience with stuff like that

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,19 +12,20 @@ Feel free to join the IRC channel: #emersion on irc.libera.chat.
 
 ## Running
 
+`mako` requires an active session bus to function properly. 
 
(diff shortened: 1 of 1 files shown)


emersion:
Hm… I'm not sure how this patch improves the readability?

o0nd7ots:
It added a separate section for running with multiple notification daemons and clearly separated systemd section from nonsystemd.

I think this makes the text more clear in a way that reader finds what they want quickly.

-------- Original Message --------

PR #509, opened 2024-03-22
implement mininum width for notification

### LINKED ISSUE #69: Let us set max width. *feature request*
Hi,

I use both long and short notification in terms of line lenght. I can set width but that's looking ugly for short notification.

See screenshot of what I mean, or maybe honoring the line break and calculate the lenght of the longest line? 

2018-10-01-232102_grim
2018-10-01-232201_grim

apprehensions:
fixes #69

### DIFF
--- a/config.c
+++ b/config.c
@@ -68,6 +68,7 @@ void finish_config(struct mako_config *config) {
 }
 
 void init_default_style(struct mako_style *style) {
+	style->min_width = 300;
 	style->width = 300;
 	style->height = 100;
 
@@ -217,6 +218,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	}
 
 	// Now on to actually setting things!
+
+	if (style->spec.width) {
+		target->min_width = style->min_width;
+		target->spec.min_width = true;
+	}
 
 	if (style->spec.width) {
 		target->width = style->width;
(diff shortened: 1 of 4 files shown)


apprehensions:
@emersion ?

apprehensions:
@emersion ?

apprehensions:
@emersion ?

config.c
```diff
@@ -68,6 +68,7 @@ void finish_config(struct mako_config *config) {
 }
 
 void init_default_style(struct mako_style *style) {
+	style->min_width = 300;
```
emersion, line 71: Sounds like this default would break existing config files setting `width` to something other than 300.

emersion (commented):
I can't really wrap my head around what this is doing, but it seems like the existing `width` setting is now ignored?

The original issue is about `max-width`, and this PR seems to be implementing something else?

apprehensions:
> The original issue is about max-width, and this PR seems to be implementing something else?

Author wrote:
> I use both long and short notification in terms of line lenght. I can set width but that's looking ugly for short notification.

In this PR, `width` will act as the maximum width.

config.c
apprehensions: Thanks for noticing, I'll add the ability to disable the feature if the minimum width is 0.

PR #521, opened 2024-06-07
Add reference to notify-send manual to the SEE ALSO 

RoadRoller01:
Add reference to notify-send manual to the SEE ALSO handy for new users


emersion:
But `notify-send` is part of libnotify, not in any of mako's dependencies…

RoadRoller01:
is there another way to send notifications with mako?, because look if some one wanted to test his configuration how will he do it, there is no information how to do in all manuals

PR #524, opened 2024-06-23
Implement an include option

### LINKED ISSUE #523: Implement an include option
I share my dotfiles between several machines and sometimes it's very convenient to have a common config file that can load other files that may be machine-dependent (e.g. scale factor or font size). The machine-dependent configs may belong to another repository or even be just local files.

Similar features exist in other projects, for example:
* `include` command in [sway](https://man.archlinux.org/man/sway.5#COMMANDS).
* `include` option in [foot](https://man.archlinux.org/man/foot.ini.5.en#SECTION:_main).
* `include` directive in [kitty](https://man.archlinux.org/man/kitty.conf.5#Overview).

I would suggest an `include` implementation closest to `foot`:
* The path must be an absolute path, or start with ~/.
* Multiple include directives are allowed, but only one path per directive.
* Nested imports are allowed.

I may in fact have a chance to look into it later this or next month, just want to double-check that this proposal makes sense. @emersion what do you think?

kupospelov:
Fixes #523

### DIFF
--- a/config.c
+++ b/config.c
@@ -486,6 +486,25 @@ bool apply_superset_style(
 	return true;
 }
 
+static char *expand_config_path(const char *path) {
+	if (strncmp(path, "/", 1) == 0) {
+		return strdup(path);
+	}
+
+	if (strncmp(path, "~/", 2) != 0) {
+		fprintf(stderr, "Config path must start with / or ~/\n");
+		return NULL;
+	}
+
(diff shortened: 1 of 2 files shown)


emersion (approved):
Looks good to me!

kupospelov:
Can this be merged for the next release?

emersion:
Thanks!

PR #526, opened 2024-07-22
Implement `makoctl count` script command with variants

fidelicura:
This PR adds new command to `makoctl` as `makoctl count`. It allows a user to get number of notifications that are currently in history or in a waiting state. It may be useful for software producing a lot of notifications (e.g. single-point DBs) or any other use case where notifications are used as counter for something. The result looks like this:
```
{
        "type" : "u",
        "data" : [
             5
        ]
}
```
...where `5` is the number of notifications in the history or on the screen (_waiting_) at the moment.

See also original idea from 2017 in [dunst](https://github.com/dunst-project/dunst). It is the same sub-command that was [initially named `status`, and now named `count`](https://github.com/dunst-project/dunst/commit/4d74c7b46e7a1e9660f5f116947bdbd09ce8ead8): [issue](https://github.com/dunst-project/dunst/issues/445), [commit](https://github.com/dunst-project/dunst/commit/137361a95dbf4064668f32d4872fa3093e2c1db3).

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -4,6 +4,7 @@ usage() {
 	echo "Usage: makoctl <command> [options...]"
 	echo ""
 	echo "Commands:"
+	echo "  count [waiting|history]        Show the number of notifications"
 	echo "  dismiss [-n id]                Dismiss the notification with the"
 	echo "                                 given id, or the last notification"
 	echo "                                 if none is given"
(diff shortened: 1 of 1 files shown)


emersion:
Do we really need to add a new D-Bus method here? Couldn't we use the same as `makoctl list` and use jq to count the number of entries?

fidelicura:
> Do we really need to add a new D-Bus method here? Couldn't we use the same as `makoctl list` and use `jq` to count the number of entries?

Now I don't think we need a new D-Bus method, so no. I did it as a script command in the latest force-push, but I think it is not that necessary anymore if the user can just use `jq` for this purpose. I won't be upset if you just close this PR.

emersion:
Yeah, I think I'd rather have the user use `busctl` + `jq` directly.

Issue #454, opened 2023-02-09
Icon border-radius


ryan-griffin:
I think it would be a nice feature to be able to change the border radius of the icon.

CyanChanges:
when will #528 get merged?

Frestein:
We can close it.

PR #528, opened 2024-08-12
Feature: icon-border-radius

Samferos:
Solves #454 ~~though I'm concerned about the fact that I have duplicated code with the `rounded_square` function in `icon.c` where it basically achieves the same purpose as the `set_rounded_rectangle` in `render.c`.~~

### DIFF
--- a/config.c
+++ b/config.c
@@ -100,6 +100,7 @@ void init_default_style(struct mako_style *style) {
 #endif
 	style->max_icon_size = 64;
 	style->icon_path = strdup("");  // hicolor and pixmaps are implicit.
+	style->icon_border_radius = 0;
 
 	style->font = strdup("monospace 10");
 	style->markup = true;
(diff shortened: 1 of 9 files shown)


Samferos:
not only that, the cmdline argument for the icon-border-radius isn't working
only works through config file right now

Samferos:
Comparison :
`icon-border-radius=0` 
image
`icon-border-radius=20`
image

emersion (commented):
Apart from this nit, looks good!

icon.c
```diff
@@ -294,7 +294,7 @@ struct mako_icon *create_icon(struct mako_notification *notif) {
 #endif
 
 void draw_icon(cairo_t *cairo, struct mako_icon *icon,
-		double xpos, double ypos, double scale) {
+	double xpos, double ypos, double scale) {
```
emersion, line 297: Nit: continuation lines are indented with two tabs, to make the difference with the function body clear.

Samferos: whoops
i forgot to rollback that file

emersion:
We could, or we could also leave this for later until someone needs it. It's a backwards-compatible change.

(Ideally all commits would be squashed together)

Samferos:
Should be good to go,
I just added the autocompletion cli argument I forgot to add and squashed my commits.

emersion (approved):
Thank you!

PR #530, opened 2024-08-22
fix(makoct): Ensure same notif id is used across external call in `makoctl menu`

ErynGalen:
Consider the following:
The first notification has action `act1`. The user starts `makoctl menu` with no id specified to get actions from this notifications. While dmenu is open, a new notification appears. When the user confirms `act1` in dmenu, this action previously got applied to this new notification, which potentially has completely different actions/semantics.

Now the behaviour is that the action gets sent to the notification from which it was obtained (this is consistent with behaviour when `-n` is passed)

### Notes
The new behaviour could already be replicated on the user's side, but there doesn't seem to be a point in keeping the previous behaviour. The previous behaviour can be achieved by passing `-n 0`

### DIFF
--- a/makoctl
+++ b/makoctl
@@ -112,11 +112,17 @@ case "$1" in
 	require_jq
 	if [ $# -gt 1 ] && [ "$1" = "-n" ]; then
 		id="$2"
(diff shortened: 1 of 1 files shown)


emersion:
makoctl has been rewritten in C, and no longer has this issue.

PR #532, opened 2024-08-24
 implement notification time freezing 

chayleaf:
Depends on #531


apprehensions:
Shouldn't this use a makoctl command instead of a config file change?

PR #533, opened 2024-08-24
gitignore: add clangd cache dir

chayleaf:
another example https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/4918


emersion:
Sorry, I'd rather not add tool-specific entries here.

For a solution, see https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

PR #534, opened 2024-08-24
add on-long-touch option

chayleaf:
There are much less input options when using a touchscreen, so hopefully this should somewhat bridge the gap in experience.

### DIFF
--- a/doc/mako.5.scd
+++ b/doc/mako.5.scd
@@ -57,9 +57,22 @@ Supported options:
 	Default: dismiss
 
 *on-touch*=_action_
-	Performs the action when tapped via a touch device.
+	Performs the action when tapped via a touch device if the tap
+	duration is less than *long-press-duration*.
 
 	Default: dismiss
+
+*on-long-touch*=_action_
+	Performs the action when tapped via a touch device if the press
+	duration is greater or equal to *long-press-duration*.
+
+	Default: invoke-default-action
+
+*long-press-duration*=_time_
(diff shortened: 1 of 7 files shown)


wayland.c
```diff
@@ -144,7 +158,7 @@ static void touch_handle_up(void *data, struct wl_touch *wl_touch,
 	wl_list_for_each(notif, &state->notifications, link) {
 		if (hotspot_at(&notif->hotspot, seat->touch.pts[id].x, seat->touch.pts[id].y)) {
 			struct mako_surface *surface = notif->surface;
-			notification_handle_touch(notif, &ctx);
+			notification_handle_touch(notif, &ctx, time - seat->touch.pts[id].time);
```
emersion, line 161: If we already have a timer, I don't think we need to also handle the timeout here?

doc/mako.5.scd
```diff
@@ -57,10 +57,23 @@ Supported options:
...
+	Performs the action when tapped via a touch device if the press
+	duration is greater or equal to *long-press-duration*.
+
+	Default: invoke-default-action
+
+*long-press-duration*=_time_
+	Specifies the cutoff time (in milliseconds) for a press to be
+	considered a long press.
+
+	Default: 500
```
emersion, line 75: I'd prefer not to expose a config option for this for now. We can always add in the future if needed, but if there's no clear use-case for it, I prefer not to add something which might tie our hands later.

PR #538, opened 2024-09-12
README: Fixed broken freedesktop link. Linted with markdownlint

Ac5000:
This pull request is mostly to fix the broken link to the Freedesktop specification, since the original link appears to be gone/broken. (Line 69/70)

I use [markdownlint](https://github.com/DavidAnson/markdownlint) and it happened to catch/fix more than I intended... (Lint on save got me...)

However, the other changes it made does push the README.md closer to the [CommonMark](https://spec.commonmark.org/current/) specification if you are OK with leaving them in.

From what I can see visually in the GitHub previewer, nothing changed visually from the current version. But, I understand if you don't wish to change the other formatting.

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,5 +67,5 @@ See the [faq section in the wiki](https://github.com/emersion/mako/wiki/Frequent
 
 MIT
 
-[spec]: https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html
+[spec]: https://specifications.freedesktop.org/notification-spec/latest/
 [basu]: https://github.com/emersion/basu
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ Feel free to join the IRC channel: #emersion on irc.libera.chat.
 
 ## Running
 
-
 `mako` will run automatically when a notification is emitted. This happens via
 D-Bus activation, so you don't really need to explicitly start it up (this also
 allows delaying its startup time and speed up system startup).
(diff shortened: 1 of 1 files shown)


emersion:
Hm, I'm not exactly sure why the exclamation mark has been removed, or why the asterisks have been replaced with dashes (the former are valid CommonMark)...

Rest LGTM!

Ac5000:
@emersion Apologies for the delayed response. 

I am going to try to do this again and make note of the references it gave for the changes it made. This way we both can understand why it made them. I don't believe I have any special configuration with markdownlint, so I was under the assumption it was only making changes aligned with CommonMark, but I may be incorrect.

While I work on getting that done, would you prefer I close this pull and open a new one with only the broken link fixed?

emersion:
Sure! Or you can undo the linter changes in this PR.

Ac5000:
@emersion Reverted the linter changes and recommitted to this pull as requested.

As promised, I also noted why it made the linting changes so we may both learn.
- Line 15 removed - [MD012](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md012---multiple-consecutive-blank-lines) - Multiple consecutive blank lines
- Lines 40-48 - [MD004](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004---unordered-list-style) - Unordered list style - Asterisks instead of dashes.
    - This was triggered because lines 23-26 used dashes so it's triggering the **consistency** check. 
    - Had lines 23-26 used asterisks it would not have cared.
- Line 62 - [MD026](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004---unordered-list-style) - Trailing punctuation in heading.
    - They explicitly call out `!` as something you can remove from the list since it's pretty common.
    - Commonmark doesn't mention punctuation for ATX headings so this must be more a preferential thing?

emersion:
Thanks!

PR #552, opened 2025-01-17
dbus: add Modes property

gkleen:
This adds a dbus property `Modes` and invalidates it on changes.
Dbus clients can subscribe to these invalidations to react in real time.


emersion (approved):
Thanks!

Issue #512, opened 2024-03-27
No notification display after monitor hotplug


Emantor:
Running latest mako git, sometimes after a hotplug notifications are no longer displayed. The service is still running and seems to receive notifications, but no longer creates a visible surface.

My hotplug configurations mostly disable my internal laptop display.

Emantor:
Hm, probably because `handle_set_mode()` calls `reapply_config()` which in turn removes the notification surfaces and sets them all to dirty which will recreate them in the next cycle.

nermolov:
Can confirm I have this issue on mako 1.9.0 and sway 1.10 after connecting to an external display and disabling my internal display.

> I noticed that `makoctl set-mode default` resolves this problem. So apparently changing screen configuration or disabling the last screen enables some kind of dnd mode?

This does also fix things for me, but I can't seem to figure out how to get sway to automatically run this command on display changes...

Emantor:
> Can confirm I have this issue on mako 1.9.0 and sway 1.10 after connecting to an external display and disabling my internal display.
> 
> > I noticed that `makoctl set-mode default` resolves this problem. So apparently changing screen configuration or disabling the last screen enables some kind of dnd mode?
> 
> This does also fix things for me, but I can't seem to figure out how to get sway to automatically run this command on display changes...

Should be doable via `kanshi`, it can run commands when the output configuration changes.

I'll have some free time in december to investigate this further.

zickgraf:
> This does also fix things for me, but I can't seem to figure out how to get sway to automatically run this command on display changes...

In case it is useful for you:
I'm working around this issue by running the following loop in the background (in bash):
```
while true; do
	swaymsg -t subscribe '["output"]';
	sleep 3;
	makoctl set-mode default;
done
```
Probably this is not the most efficient solution but it gets the job done :D

mbachry:
I'm running my own mako build with a `reapply_config` call added to `destroy_output` function. Seems to be doing the job so far. The actual fix probably needs more fine grained surface/frame/etc invalidation, but on the other hand why bother when destroying output is such a rare event.

Here's my patch:

```
diff --git a/dbus/mako.c b/dbus/mako.c
index 4c95881..1323fb3 100644
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -288,7 +288,7 @@ static int handle_list_history(sd_bus_message *msg, void *data,
  *    This will also (re)create all surfaces we need in the new config.
  * 3. Start the redraw events.
  */
[...]

PR #558, opened 2025-02-22
Fix use after free when closing notifications

drunkenblueberry:
This might be related to https://github.com/emersion/mako/issues/512, as the problem I saw had similar symptoms. But at some point, I started to see segfaults each time I closed a notification, regardless of whether I (un)plugged monitors.

The crux of the issue is that closing a notification frees it, and in some cases the notification was being used after being closed.


emersion (approved):
Good catch, thanks!

PR #559, opened 2025-02-23
Add max-urgency option

ThinkChaos:
Allow limiting notification urgency to a given maximum.

I'm unfortunately forced to use MS Teams for work, and basically all its notifications are marked as urgent (at least when using the website). This change allows me to cap that:
```ini
[app-name=Chromium]
max-urgency=normal
```
(I use a dedicated browser for work, so app-name is enough filtering.)


emersion:
I'd prefer a more general approach to this problem, although it is more work: https://github.com/emersion/mako/issues/566

PR #564, opened 2025-02-28
Unify Dismiss* methods and add --no-history to `makoctl`

qiu-x:
This PR addresses the D-Bus API clean-up mentioned in https://github.com/emersion/mako/pull/495#issuecomment-2067702141 and also makes stuff more coherent. It also adds a `--no-history` flag to `makoctl dismiss` so #495 could also be closed if this gets into master.

### DIFF
--- a/dbus/mako.c
+++ b/dbus/mako.c
@@ -14,76 +14,72 @@
 static const char *service_path = "/fr/emersion/Mako";
 static const char *service_interface = "fr.emersion.Mako";
 
-static int handle_dismiss_all_notifications(sd_bus_message *msg, void *data,
-		sd_bus_error *ret_error) {
-	struct mako_state *state = data;
-
-	close_all_notifications(state, MAKO_NOTIFICATION_CLOSE_DISMISSED);
-	struct mako_surface *surface;
-
-	wl_list_for_each(surface, &state->surfaces, link) {
-		set_dirty(surface);
-	}
-
-	return sd_bus_reply_method_return(msg, "");
-}
-
-static int handle_dismiss_group_notifications(sd_bus_message *msg, void *data,
-		sd_bus_error *ret_error) {
-	struct mako_state *state = data;
-
-	if (wl_list_empty(&state->notifications)) {
-		goto done;
-	}
-
-	struct mako_notification *notif =
-		wl_container_of(state->notifications.next, notif, link);
-
-	struct mako_surface *surface = notif->surface;
-	close_group_notifications(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED);
-	set_dirty(surface);
-
-done:
-	return sd_bus_reply_method_return(msg, "");
-}
-
-static int handle_dismiss_last_notification(sd_bus_message *msg, void *data,
-		sd_bus_error *ret_error) {
-	struct mako_state *state = data;
-
-	if (wl_list_empty(&state->notifications)) {
-		goto done;
-	}
-
-	struct mako_notification *notif =
-		wl_container_of(state->notifications.next, notif, link);
-	struct mako_surface *surface = notif->surface;
-	close_notification(notif, MAKO_NOTIFICATION_CLOSE_DISMISSED, true);
-	set_dirty(surface);
-
-done:
-	return sd_bus_reply_method_return(msg, "");
-}
-
-static int handle_dismiss_notification(sd_bus_message *msg, void *data,
+static int handle_dismiss(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct mako_state *state = data;
 
 	uint32_t id = 0;
-	int dismiss_group = 0;
-	int ret = sd_bus_message_read(msg, "ub", &id, &dismiss_group);
+	int group = 0;
+	int all = 0;
+	int history = 1; // Keep history be default
+
+	int ret = sd_bus_message_enter_container(msg, 'a', "{sv}");
 	if (ret < 0) {
 		return ret;
+	}
+
+	while (true) {
+		ret = sd_bus_message_enter_container(msg, 'e', "sv");
+		if (ret < 0) {
+			return ret;
+		} else if (ret == 0) {
+			break;
+		}
+
+		const char *key = NULL;
+		ret = sd_bus_message_read(msg, "s", &key);
+		if (ret < 0) {
+			return ret;
+		}
+
+		if (strcmp(key, "id") == 0) {
+			ret = sd_bus_message_read(msg, "v", "u", &id);
+		} else if (strcmp(key, "group") == 0) {
+			ret = sd_bus_message_read(msg, "v", "b", &group);
+		} else if (strcmp(key, "history") == 0) {
+			ret = sd_bus_message_read(msg, "v", "b", &history);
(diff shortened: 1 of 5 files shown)


qiu-x:
Hi, @emersion,
I know you're probably pretty busy with wlroots and your day job. Nonetheless, I would really appreciate it if you could take a look at my changes before the next release of mako. I really like how slim and light the project is compared to dunst and these changes might improve the flow of people using the history feature. Thanks for all of your amazing work!

dbus/mako.c
```diff
@@ -468,10 +470,7 @@ void emit_modes_changed(struct mako_state *state) {
 
 static const sd_bus_vtable service_vtable[] = {
 	SD_BUS_VTABLE_START(0),
-	SD_BUS_METHOD("DismissAllNotifications", "", "", handle_dismiss_all_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("DismissGroupNotifications", "", "", handle_dismiss_group_notifications, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("DismissLastNotification", "", "", handle_dismiss_last_notification, SD_BUS_VTABLE_UNPRIVILEGED),
-	SD_BUS_METHOD("DismissNotification", "ub", "", handle_dismiss_notification, SD_BUS_VTABLE_UNPRIVILEGED),
+	SD_BUS_METHOD("Dismiss", "a{sv}", "", handle_dismiss, SD_BUS_VTABLE_UNPRIVILEGED),
```
emersion, line 473: Nit: I'd prefer to keep "notification" in the name. Perhaps "DismissNotifications" since it might dismiss one or more?

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
-		sd_bus_error *ret_error) {
-	struct mako_state *state = data;
+		ret = sd_bus_message_exit_container(msg);
+		if (ret < 0) {
+			return ret;
+		}
+	}
 
-	if (wl_list_empty(&state->notifications)) {
+	// These don't make sense together
```
emersion, line 70: Maybe we should return an error instead of doing nothing?

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
+	if (all && group)
+	{
```
emersion, line 72: Style: `{` should be on the previous line

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
+		}
 
-done:
-	return sd_bus_reply_method_return(msg, "");
-}
+		if (strcmp(key, "id") == 0) {
+			ret = sd_bus_message_read(msg, "v", "u", &id);
+		} else if (strcmp(key, "group") == 0) {
+			ret = sd_bus_message_read(msg, "v", "b", &group);
+		} else if (strcmp(key, "no-history") == 0) {
```
emersion, line 53: Instead of using negative options, I'd prefer to define a positive option and default it to true:

```c
int history = 1;
if (strcmp(key, "history") == 0) {
	ret = sd_bus_message_read(msg, "v", "b", &history);
}
```

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
+static int handle_dismiss(sd_bus_message *msg, void *data,
 		sd_bus_error *ret_error) {
 	struct mako_state *state = data;
 
-	close_all_notifications(state, MAKO_NOTIFICATION_CLOSE_DISMISSED);
-	struct mako_surface *surface;
-
-	wl_list_for_each(surface, &state->surfaces, link) {
-		set_dirty(surface);
+	if (wl_list_empty(&state->notifications)) {
```
emersion, line 21: I think I'd prefer to remove this check, so that the function behaves the same when there are notifications and when there are none.

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
-		sd_bus_error *ret_error) {
-	struct mako_state *state = data;
-
-	uint32_t id = 0;
-	int dismiss_group = 0;
-	int ret = sd_bus_message_read(msg, "ub", &id, &dismiss_group);
-	if (ret < 0) {
-		return ret;
-	}
+	bool add_to_hisotry = !no_history;
```
emersion, line 76: Typo: "history"

dbus/mako.c
```diff
@@ -14,82 +14,84 @@
...
-	struct mako_state *state = data;
+		ret = sd_bus_message_exit_container(msg);
+		if (ret < 0) {
+			return ret;
+		}
+	}
 
-	if (wl_list_empty(&state->notifications)) {
+	// These don't make sense together
+	if (all && group)
```
emersion, line 71: `id` is also incompatible with `all` I think?

makoctl.c
```diff
@@ -109,16 +111,63 @@ static int run_dismiss(sd_bus *bus, int argc, char *argv[]) {
...
 		}
 	}
 
-	if (all) {
-		return call_method(bus, "DismissAllNotifications", NULL, "");
-	} else {
-		return call_method(bus, "DismissNotification", NULL, "ub", id, (int)group);
+	if (all && group)
+	{
+			fprintf(stderr, "-a and -g cannot be used together\n");
```
emersion, line 124: Style: indentation is off here

makoctl.c
```diff
@@ -109,16 +111,63 @@ static int run_dismiss(sd_bus *bus, int argc, char *argv[]) {
...
+			return -EINVAL;
+	}
+
+  char types[6] = "a{sv}";
```
emersion, line 128: Style: indentation should use tabs

emersion (commented):
Thanks for the PR! Here are a few comments.

dbus/mako.c
qiu-x: right, I didn't think of that

dbus/mako.c
qiu-x: Should it still stay as `--no-history` in makoctl or do you prefer it to be a positive option there as well?
I used a negative option to be consistent with the `--no-history` option for the `dismiss` action in the config.

emersion: Yeah, I think for CLI arguments we can stick to negative options, because it's more consistent with the status quo and because boolean CLI options don't take a value.

qiu-x:
Thanks for the feedback! I adjusted the changes according to your comments. Please let me know if any further modifications are necessary. I can squash the all the changes into one commit if all looks good.

dbus/mako.c
qiu-x: I changed it to `DismissNotifications`

notification.c
```diff
@@ -167,12 +167,13 @@ struct mako_notification *get_tagged_notification(struct mako_state *state,
 }
 
 void close_group_notifications(struct mako_notification *top_notif,
-	       enum mako_notification_close_reason reason) {
+	       enum mako_notification_close_reason reason,
+			   bool add_to_history) {
```
emersion, line 171: One last detail: the indentation here is a bit off, we use 2 tabs for function arguments.

emersion (commented):
LGTM apart from this last detail! Feel free to squash.

emersion (approved):
Thanks!

PR #573, opened 2025-03-23
build: optionally install systemd service

LaserEyess:
This is based off of how systemd support is handled in [`xdpw`](https://github.com/emersion/xdg-desktop-portal-wlr/blob/master/meson.build#L89-L102). 

I believe this is a genuinely useful feature, and that it does not interfere with non-systemd based distros/OSes in any meaningful way. Downstream distros have been carrying support for this for some time:

 - [Fedora]( https://src.fedoraproject.org/rpms/mako/blob/rawhide/f/add-systemd-service-dbus.patch)
 - [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/mako/-/blob/main/systemd-dbus-activation.patch?ref_type=heads)
 - [Debian](https://sources.debian.org/src/mako-notifier/1.10.0-1/debian/mako-notifier.install/)

Probably others, too. These were just the first three I checked.


emersion:
FWIW, this was dropped in https://github.com/emersion/mako/pull/338.

LaserEyess:
It looks like that was more about the apparmor support than the systemd service? From a quick check I did not see a lot of issues about the systemd service that were distro specific, but I could be missing some issues.

Given that many distros seem to be carrying this patch around, and that there have not been a lot of (any?) bug reports specific to those patches, I think this is still maintainer friendly. Since I actively use mako with systemd support I would also volunteer to triage systemd bugs, fwiw.

Issue #550, opened 2025-01-07
streaming (waybar) support


anarcat:
It would be nice to have a way to "stream" notifications out so that another app can get updates when notifications land.

Right now, I can find out the list of pending notifications with `mako list`. I can make a custom waybar panel, for example, that will count the number of such entries with something like: `makoctl list | jq '.data[] | length'`, and I could integrate that in waybar.

But then I need to refresh this all the time, and that's costly (say, on battery), as I have to wakeup all the time. Meh.

A better way would be to have some sort of streaming interface where a caller could hook into mako and get pushed updates. The way [sway-notification-center](https://github.com/ErikReider/SwayNotificationCenter) works in that regard is interesting. Look at the output of the `--subscribe-waybar` flag:

```
> swaync-client --subscribe-waybar
{"text": "0", "alt": "none", "tooltip": "", "class": "none"}
```

Now, that output is kind of specific to waybar of course: this is a JSON object formatted in a way that Waybar will like. But this could be anything you like... The blobs in the `.data` field of `makoctl list` would be fine, as an intermediate filter could parse those and generate the correct waybar output. 

The key here is to have mako send those updates over a pipe somehow, instead of the callers pulling those in.

Right now, my mako integration in waybar is substandard, as the message count doesn't automatically update. But it's the best I could do with the current interface, it would be great to have a push system to improve on this!

```json
    "custom/dnd": {
        "interval": "once",
        "return-type": "json",
        "format": "{}{icon}",
        "format-icons": {
            "default": "",
            "dnd": ""
        },
        "comment": "this hack is from https://github.com/manjaro-sway/desktop-settings/pull/121",
        "on-click": "makoctl mode | grep 'do-not-disturb' && makoctl mode -r do-not-disturb || makoctl mode -a do-not-disturb; pkill -RTMIN+11 waybar",
        "on-click-right": "makoctl dismiss --all ; pkill -RTMIN+11 waybar",
        "exec": "printf '{\"text\": \"%s\", \"alt\":\"%s\",\"tooltip\":\"mako notification mode: %s\"}' $(makoctl list | jq '.data[] | length') $(makoctl mode | grep -q 'do-not-disturb' && echo dnd || echo default) $(makoctl mode | tail -1)",
        "signal": 11
    },
```

I *suspect* something around dbus would be needed to implement this, but I'm not familiar enough with dbus (let alone mako's implementation) to figure this out..

Thanks!

emersion:
[...]

Issue #522, opened 2024-06-10
ERROR: Neither source directory 'build' nor build directory None contain a build file meson.build


xxxtrashcion:
Installed all dependencies but running "meson build" gives me this.

emersion:
Please ask questions on IRC, not on the issue tracker.

PR #578, opened 2025-04-01
Add an on-closed event 

Queuecumber:
In some cases (for example, informing an external monitor like waybar) it can be useful to peform an action when a notification is closed/dismissed. This PR adds a `on-closed` event to allow mako users to take such an action. 

For example, to signal waybar to reload its state, the user could add:

```
on-closed=exec pkill -SIGRTMIN+5 waybar
```

### DIFF
--- a/config.c
+++ b/config.c
@@ -154,6 +154,7 @@ void finish_style(struct mako_style *style) {
 	finish_binding(&style->button_bindings.right);
 	finish_binding(&style->touch_binding);
 	finish_binding(&style->notify_binding);
+	finish_binding(&style->closed_binding);
 	free(style->icon_path);
 	free(style->font);
 	free(style->format);
@@ -398,6 +399,11 @@ bool apply_style(struct mako_style *target, const struct mako_style *style) {
 	if (style->spec.notify_binding) {
 		copy_binding(&target->notify_binding, &style->notify_binding);
 		target->spec.notify_binding = true;
+	}
+
+	if (style->spec.closed_binding) {
+		copy_binding(&target->closed_binding, &style->closed_binding);
+		target->spec.closed_binding = true;
 	}
 
 	return true;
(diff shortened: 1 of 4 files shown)


emersion:
I don't think users should configure mako to reload the state of indicators such as waybar. Instead, I think mako should provide APIs to allow indicators to subscribe to state changes, e.g. with a D-Bus event.

Queuecumber:
Isn't the notification API already a dbus API?

emersion:
mako has two D-Bus APIs: the standard D-Bus one for applications submitting notifications, and a mako-specific one for makoctl and other tools. We already expose the current list of notifications there. We could add an event fired when a notification is added or dismissed.

Queuecumber:
This sounds like something that would replace the current `on-X` config system

emersion:
No: `on-X` is meant to be used by users, while D-Bus events are meant to be used by other tools interacting with mako.

emersion:
See https://github.com/emersion/mako/pull/552 for an example of a D-Bus event approach, for modes.

Queuecumber:
I'm not sure I understand the distinction, at least not in this case. 

Certainly if I was developing an application that depended on Mako then listening for DBus notifications is the way to go, it's a more efficient and cleaner design. 

But that isn't whats happening with Waybar, it doesn't have first-class support for any notification system so you have to configure it to call scripts which interrogate Mako (or whatever you are using), and your options are either poll or use unix signals to coordinate the state updates.

Maybe there is a way to do that with DBus using a continuously running loop that sits on the bus and waits for notifications then emits JSON (I think this is supported by Waybar? Maybe worth a shot...) but either way I am approaching this as a user, at least for now. 

Also I don't necessarily want to derail this too much with discussions about Waybar, I think the on-dismissed notification could be useful for other things and it feels more like feature parity with the `on-notify` event. The fact that it unblocks some bar users is incidental.

Queuecumber:
Let me also look at #522 and see if there's a similarly easy way to add a DBus event for notify and dismiss since I'm poking around anyway

Queuecumber:
Please see #579 

This also turned out to be relatively simple including the Waybar integration (which I will follow up on in #550) 

That MR may supersede this one or compliment it depending on your preference

Issue #541, opened 2024-09-20
Transient Notification


CoolnsX:
As a long dunst user, I have decided to try mako.

I have configured in such a way, that it almost look like my old configured dunst.

Only feature I will miss is 'Transient Notification', which when emit by notify-send via  -e args, doesn't store notification in history. 

I do use history a lot, and I don't want to lose my important notifications just because my mako history is filled with notify-send garbages.

I have lots of shell scripts that runs at login and are part of my workflow.

If anyone have any workaround to do this, thanks in advance.

emersion:
Would it make sense to add a new configuration option to selectively disable history based on criteria?

Kind of related: https://github.com/emersion/mako/pull/477

emersion:
Hm -- the D-Bus protocol has a "transient" hint - I wonder if that should skip history.

CoolnsX:
@emersion this is the dunst PR for it --
https://github.com/dunst-project/dunst/issues/310

CoolnsX:
you can follow the same way, like ignore transient hint, and only do when it's explicitly set in config.

CoolnsX:
thanks in advance

PR #582, opened 2025-04-21
Handle transient notifications (#541)

grmat:
Draft implementation for #541, implements handling of the "transient" hint according to the [Desktop Notifications Specification](https://specifications.freedesktop.org/notification-spec/latest-single/#hints).

The `-e`/`--transient` flag of `notify-send` (libnotify) allows a simple test:

```
notify-send -ea mailer -i email "from <test@example.org>" "Subject"
```

While users are still able to explicitly put notifications in the history when calling `makoctl dismiss` (without `--no-history`) or using the according binding (`MAKO_BINDING_DISMISS` instead of `MAKO_BINDING_DISMISS_NO_HISTORY`).

---

Second commit is some kind of a clean-up, as it doesn't make much sense, IMHO to maintain both, a transient member and a (no-)history parameter.

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -92,6 +92,11 @@ bool match_criteria(struct mako_criteria *criteria,
 		return false;
 	}
 
+	if (spec.transient &&
+			criteria->transient != notif->transient) {
+		return false;
+	}
+
 	if (spec.urgency &&
 			criteria->urgency != notif->urgency) {
 		return false;
@@ -480,6 +485,7 @@ struct mako_criteria *create_criteria_from_notification(
 	criteria->app_icon = strdup(notif->app_icon);
 	criteria->actionable = !wl_list_empty(&notif->actions);
 	criteria->expiring = (notif->requested_timeout != 0);
+	criteria->transient = notif->transient;
 	criteria->urgency = notif->urgency;
 	criteria->category = strdup(notif->category);
 	criteria->desktop_entry = strdup(notif->desktop_entry);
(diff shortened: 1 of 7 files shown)


emersion:
Hm, the way I've read the spec, the "transient" bit meant that the notification is not stored on disk (and restored), but history sounded unrelated.

balsoft:
@emersion the spec is as follows:

> `transient`: When set the server will treat the notification as transient and by-pass the server's persistence capability, if it should exist. 

Where the persistence capability is defined as follows:

> `persistence`:  The server supports persistence of notifications. Notifications will be retained until they are acknowledged or removed by the user or recalled by the sender. The presence of this capability allows clients to depend on the server to ensure a notification is seen and eliminate the need for the client to display a reminding function (such as a status icon) of its own. 

I don't think it has anything to do with storing it on disk.

The way this persistence is usually implemented is that there's a pop-up when the notification is first emitted, and then if it times out the notification is visible when the user accesses the "notification drawer", where they can be dismissed or actioned on. `resident` notifications are those which don't disappear even if the user does some action to them while they are in the pop-up state, and need to be dismissed explicitly from the drawer. `transient` notifications are those which appear once as a pop-up and don't show up in the drawer at all.

Since mako doesn't have a "notification drawer" per se, the closest analogue we can come up with is that the usual notification display is the "pop-up", and the notification history is the "drawer". It seems like this PR implements this behavior perfectly (at least for transient notifications).

From my experience this is also how the hint is usually used by applications: to signal notifications which are simply pop-ups and don't need to be recalled later on.

PR #583, opened 2025-05-07
Add ability to scroll through group notifications

tiagoporsch:
Hello,

I searched around a little but could not find the means to achieve the functionality that I was looking for, so I've added a way to scroll through group notifications using the mouse wheel.

I've tested it around a bit and it seems to work fine, but I understand that the way I've implemented it is probably not the best, so any further contribution would be much appreciated.

To-do:
- Add an option to disable, or make this functionality opt-in
- Make this work using a touchpad

Let me know what you think about this!

### DIFF
--- a/include/notification.h
+++ b/include/notification.h
@@ -30,6 +30,7 @@ struct mako_notification {
 	uint32_t id;
 	int group_index;
 	int group_count;
+	int group_focus_index;
 	bool hidden;
 
 	char *app_name;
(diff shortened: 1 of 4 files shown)


tiagoporsch:
@emersion any feedback on this idea? I'm still running this branch without any problems, and I think it's very important to be able to see more than the latest/earliest notification in a group.

Issue #581, opened 2025-04-14
Support picking up the icon to use from `desktop-entry` hint


quite:
Support for loading icon from `desktop-entry` hint would be nice. https://specifications.freedesktop.org/notification-spec/1.3/hints.html#id-1.9.7

Can libsfdo help out with location and readin the .desktop-file? https://github.com/emersion/mako/issues/555

A couple of observation in the wild, on my Arch Linux systems. The following notifications came from the element-desktop and signal-desktop packages respectively, installed from the official binary repository "extra".

```
Notification 1364: <a-matrix-channel-name>
  App name: Element
  Desktop entry: Element
  Urgency: normal
  Actions:
    default: View
Notification 1363: Signal
  App name: Signal
  Desktop entry: signal
  Urgency: normal
  Actions:
    default: View
```

/usr/share/applications/io.element.Element.desktop:
```
[Desktop Entry]
Name=Element
Comment=Feature-rich client for Matrix
Exec=/usr/bin/element-desktop %u
Terminal=false
Type=Application
Icon=io.element.Element
StartupWMClass=Element
Categories=Network;InstantMessaging;Chat;IRCClient
MimeType=x-scheme-handler/element;
```

/usr/share/applications/signal-desktop.desktop:
```
[Desktop Entry]
Type=Application
Name=Signal
Comment=Signal - Private Messenger
Comment[de]=Signal - Sicherer Messenger
Icon=signal-desktop
Exec=signal-desktop -- %u
Terminal=false
Categories=Network;InstantMessaging;
StartupWMClass=signal
MimeType=x-scheme-handler/sgnl;x-scheme-handler/signalcaptcha;
Keywords=sgnl;chat;im;messaging;messenger;security;privat;
X-GNOME-UsesNotifications=true
```

So it looks like the desktop-entry references the `Name` in the desktop-file (possibly case-insensitive, or it's just a mistake there with "signal"; i don't know what other notification daemons name of this).

emersion:
Sounds sensible to me. Yeah, I think libsfdo can help with this.

quite:
Started a TODO in the description/first "comment" above.

quite:
I looked quickly at a few different notification daemons. I found https://gitlab.xfce.org/apps/xfce4-notifyd that does support the desktop-entry hint. It seems to be opportunistic in how it interprets the desktop-entry hint value. Like, first exact match on a .desktop-file. Then case-insensitive substring match on Name= in every desktop file?

PR #584, opened 2025-05-29
Use libsfdo-based symbolic/scalable/SVG icon lookup

### LINKED ISSUE #555: Switch to libscfg for icon lookup
https://gitlab.freedesktop.org/vyivel/libsfdo/

arebaka:
This PR introduces native Freedesktop Icon Theme support in Mako by leveraging libsfdo for scalable (.svg) and high-DPI icons. Previously, Mako only searched for bitmap icons under hard-coded icon-path directories. Now, users can specify one or more icon themes (e.g. Adwaita:Tela-green-dark) in their configuration, and Mako will:

1. First attempt to resolve the notification’s icon_path through libsfdo using the configured theme names.
2. Then, if no icon is found via libsfdo, fall back to the existing icon-path lookup logic (unchanged).

Changes:

- New icon-theme config key
Accepts a colon-separated list of Freedesktop icon theme names.
Example:
```ini
icon-theme = Adwaita:Tela-green-dark
icon-path = /usr/share/icons/Papirus # fall back
max-icon-size = 48
```

- libsfdo integration
Create and configure an sfdo_icon_lookup context using the user’s icon-theme list.
Set desired icon size, output scale, and fallback theme (hicolor) via libsfdo.
Call sfdo_icon_theme_lookup() to fetch scalable and SVG icons.

- Fallback behavior
If icon-theme is unset or libsfdo lookup fails, Mako reverts to its original icon-path + glob() search logic without any changes.

- No breaking changes for users who don’t set icon-theme; existing behavior is preserved.

Testing:

- Verified that SVG and high-DPI icons are correctly resolved when icon-theme is set.
- Confirmed that bitmap (.png, .xpm) icons in icon-path still load when no matching theme icon exists.
- Manual tests across Wayland outputs with scale factors 1 and 2.

Closes #555
CC: @arebaka

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -8,6 +8,7 @@ packages:
   - pango-dev
   - gdk-pixbuf-dev
   - scdoc
+  - libsfdo
 sources:
   - https://github.com/emersion/mako
 tasks:
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -8,7 +8,7 @@ packages:
   - pango-dev
   - gdk-pixbuf-dev
   - scdoc
-  - libsfdo
+  - libsfdo-dev
 sources:
   - https://github.com/emersion/mako
 tasks:
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -8,6 +8,7 @@ packages:
   - gdk-pixbuf2
   - scdoc
   - systemd
+  - libsfdo
 sources:
   - https://github.com/emersion/mako
 tasks:
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -13,7 +13,6 @@ packages:
   - scdoc
   - wayland
   - wayland-protocols
-  - libsfdo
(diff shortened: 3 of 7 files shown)


emersion:
Thanks for the patch! Instead of maintaining two codepaths, could we get rid of the old one? Or are there some lookups we can't support with the new one?

> Need to install the libsfdo library on the build server to pass the auto checks :)

This can be done by adding new packages to `.builds/*.yml`.

emersion:
> New icon-theme config key

Seems like this is a separate, new feature - could you split it in a separate commit, or leave it for a future PR?

arebaka:
I don't know what the libsfdo package (port) is called on FreeBSD, or if it even exists.

arebaka:
@emersion review and help with the FreeBSD build so I can update the docs and resolve #581.

quite:
I had not seen this PR addressing https://github.com/emersion/mako/issues/581 that i filed. What is blocking here? Would be very nice to get this support in. <3

PR #585, opened 2025-05-29
Add min-icon-size option to enforce minimum icon dimensions

### LINKED ISSUE #464: feat: min-icon-size
there is already a max-icon-size option, can we have a min-icon-size option please?

arebaka:
This PR adds support for a new configuration option and command-line flag: min-icon-size. This setting ensures that icons in notifications are not rendered too small, especially in cases where the icon file (e.g. .svg or low-res .png) is smaller than desired.

Previously, icons were rendered at their native size. This often led to extremely small icons in notifications (e.g. 16x16px icons on high-DPI screens), making them barely visible. Now, Mako will check the dimensions of the loaded icon and, if the largest side is smaller than min-icon-size, it will reload the image using gdk_pixbuf_new_from_file_at_scale() to upscale it to the minimum size while maintaining aspect ratio.

Was:
изображение

Became:
изображение

Changes:

- New min-icon-size configuration key
Example:
min-icon-size = 32

- New command-line override: --min-icon-size=N

- Scaling logic:
After loading the icon image (via gdk_pixbuf_new_from_file()), check if its width or height is less than min-icon-size.
If so, reload the image using gdk_pixbuf_new_from_file_at_scale() with the minimum dimension.
Aspect ratio is preserved.

Benefits:

- Improves visibility and aesthetics of notifications using small icons.
- Particularly useful for .svg icons or legacy bitmap icons at low resolutions.
- Works seamlessly with both old icon-path logic and new icon-theme lookup #584 (if present).

Compatibility:

- The new min-icon-size option is optional.
- If unset, Mako retains its original icon rendering behavior.
- Fully backward compatible with existing configurations.

Testing:

- Verified behavior with small PNG icons (16x16) and scalable SVG icons.
- Confirmed that large icons are unaffected.
- Checked integration with both icon lookup paths (libsfdo and legacy glob).

Closes #464
CC: @arebaka

### DIFF
--- a/config.c
+++ b/config.c
@@ -100,6 +100,7 @@ void init_default_style(struct mako_style *style) {
 #endif
 	style->max_icon_size = 64;
 	style->icon_path = strdup("");  // hicolor and pixmaps are implicit.
+	style->icon_theme = strdup("");
 	style->icon_border_radius = 0;
 
 	style->font = strdup("monospace 10");
@@ -155,6 +156,7 @@ void finish_style(struct mako_style *style) {
 	finish_binding(&style->touch_binding);
 	finish_binding(&style->notify_binding);
 	free(style->icon_path);
+	free(style->icon_theme);
 	free(style->font);
 	free(style->format);
 	free(style->output);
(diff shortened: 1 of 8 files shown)


ClairDePrince:
Why is this not being merged?

PR #586, opened 2025-05-30
Add JSON output support to makoctl list and makoctl history

### LINKED ISSUE #580: makoctl: re-introduce JSON output
makoctl used to output a (weird) JSON when it was backed by busctl. This was useful for processing by other tools.

Users can still use busctl manually to get that JSON, but it would be easier to provide support inside makoctl. The shape of the JSON can be improved too.

arebaka:
This pull request adds support for JSON-formatted output via the -j flag for the following commands:

- makoctl list
- makoctl history

When the flag is provided, the output is serialized as JSON, making it easier to consume in scripts and external tools.

Changes:

- Added --json flag parsing to relevant CLI commands
- Implemented recursive serialization of D-Bus sd_bus_message structures into valid JSON
- Supports:
  - Basic types: string, bool, byte, int32, uint32
  - Variants (v)
  - Dictionaries (a{sv}, a{ss})
  - Arrays (a[...])
- Safely escapes strings and handles nested containers

Example usage:
```bash
makoctl list -j | jq .
makoctl history --json | jq '.[].summary'
```

Tested:
- Output is valid JSON (verified with jq)
- Handles all expected fields: summary, app-name, urgency, actions, etc
- Works with nested and complex D-Bus structures

Motivation:

This makes makoctl significantly more script- and automation-friendly, especially for use in status bars, dashboards, or notification diagnostics.

Closes #580
CC: @arebaka

### DIFF
--- a/makoctl.c
+++ b/makoctl.c
@@ -280,6 +280,207 @@ static bool is_empty_str(const char *str) {
 	return str == NULL || str[0] == '\0';
 }
 
+static void escape_and_print_json_string(const char *s) {
+    putchar('"');
+
+    for (; *s; ++s) {
+        switch (*s) {
+            case '\"': printf("\\\""); break;
+            case '\\': printf("\\\\"); break;
+            case '\b': printf("\\b");  break;
+            case '\f': printf("\\f");  break;
+            case '\n': printf("\\n");  break;
+            case '\r': printf("\\r");  break;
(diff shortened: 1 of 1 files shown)


arebaka:
I don't know what the libsfdo package (port) is called on FreeBSD, or if it even exists.

parisni:
really nice, hope json output could be restored soon on mako

Issue #211, opened 2019-12-21
custom icon


CorruptCommit:
I would like to request an option to use a custom icon something like this:

```
[summary=OMG]
background-color=#C40023
icon=/path/to/OMG.png
```

emersion:
Maybe we could add e.g. `default-icon` and `ignore-icon`, like we have for `timeout`.

vilhalmer:
`ignore-icon` would overlap a bit weirdly with the fact that you can already override `icons` within criteria sections, but I like the symmetry in spirit.

emersion:
>you can already override icons within criteria sections

Hmm, how so?

vilhalmer:
Setting `icons=false` in criteria will hide icons for matching notifications without disabling icon support globally. (Or conversely you can enable them only for certain criteria.)

emersion:
Oh right. We should probably use the same logic for icons and timeouts somehow.

vilhalmer:
I'd be fine with a breaking change to make `icons` a global-only option and then adding `ignore-icons` for criteria.

PR #587, opened 2025-05-31
Add support for override.* keys in to apply property overrides after matching criterias

### LINKED ISSUE #566: Generalized defaults/overrides for notification properties
Ref https://github.com/emersion/mako/pull/559  
Ref https://github.com/emersion/mako/issues/211  

Proposal: `override.*` keys to override properties, combined with generalized criteria to select when to override.

```ini
[!icon]
icon=default

[timeout>1000]
timeout=1000
```

Overriding interacts a bit weirdly with criteria. Do we perform another criteria pass just to apply overrides?

arebaka:
This pull request introduces a new “overrides” mechanism in Mako’s INI-style configuration. Users can now specify one or more override.* keys inside a criteria block to defer certain property assignments until after all matching criteria have been evaluated.

Closes #566
Ref #559
Ref #211
CC: @arebaka

### DIFF
--- a/config.c
+++ b/config.c
@@ -10,6 +10,7 @@
 #include "criteria.h"
 #include "string-util.h"
 #include "types.h"
+#include "styling.h"
 
 static int32_t max(int32_t a, int32_t b) {
 	return (a > b) ? a : b;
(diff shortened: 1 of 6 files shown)


ThinkChaos (commented):
Thanks for working on this, reviewing as a random mako user that's interested in such a feature, as per opening #559 a while back :)

I tried it out and unfortunately I don't think it can close #566 as is:
- applying an override doesn't cause criteria defined later in the config that now match to apply. See [this comment](https://github.com/emersion/mako/issues/566#issuecomment-2803978639) for an example config and the following comments for details on expectations.  
  This seems like the biggest issue for me cause overriding the urgency doesen't cause the notification's color to match the override, but instead use the original urgency's color, thus not solving my personal use case.
- that issue also mentions supporting criteria like `timeout>1000`. I think that could be a separate issue cause it seems like a bunch of work, but as it is, #566 is not fully resolved.

PR #600, opened 2025-09-03
Check whether the notification supports action in `makoctl invoke`

arximboldi:
When the action is not supported by the given notification, we return an error condition and report this on the terminal.

It is useful to know whether the notification was successfully invoked, for example, I have this in my `hyperland` configuration to deal with the newest notification (similar to clicking on it):
```
bind = $MOD, period, exec, makoctl invoke || makoctl dismiss
```

An alternative approach for my use-case would be to have the `InvokeAction` dismiss the notification when the `default` action is sent to an action-less notification.  The chosen approach is more versatile, as the user of `makoctl invoke` can decide what behavior to chose.

### DIFF
--- a/makoctl.c
+++ b/makoctl.c
@@ -173,35 +173,6 @@ static int run_dismiss(sd_bus *bus, int argc, char *argv[]) {
 	return call_method(bus, "DismissNotifications", NULL, types, &msg);
 }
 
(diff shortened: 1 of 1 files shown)


arximboldi:
That makes sense. I was just inspired by the `menu` command to do it this way. I don't have time to do this this week though, maybe the next one or the one after.

PR #607, opened 2025-10-27
Remove criteria for max_visible (Fix #413)

### LINKED ISSUE #413: Bug with max-visible?
While I like the man page, its description of groups/criteria is not easy to understand. For example, it says 
"...the config file may contain zero or more sections, each containing any combination of the BINDING OPTIONS and STYLE OPTIONS. "
However, even the following 2-line config gives me an error:
`[category=brightness]
max-visible=1`
This is either a bug, or It would be great if there were more examples of allowed configurations in the wiki. For example, how to define a group where notifications overwrite each other (like in brightness/audio level notifications).

FreeCap23:
Fix for https://github.com/emersion/mako/issues/413

I'm not sure if this breaks anything, but it fixes an error when trying to use max_visible under a criteria section.

After this fix exceeding the max visible number of notifications by n will still create a new notification that says "(n more)"

As a workaround, create a criteria like so to make these overflowing notifications invisible:
```
[hidden]
invisible=1
```

### DIFF
--- a/criteria.c
+++ b/criteria.c
@@ -526,12 +526,6 @@ bool validate_criteria(struct mako_criteria *criteria) {
 	copy.hidden = false;
 	bool any_but_surface = mako_criteria_spec_any(&copy);
 
-	if (criteria->style.max_visible && any_but_surface) {
-		fprintf(stderr, "Setting `max_visible` is allowed only for `output` "
-				"and/or `anchor`\n");
-		return false;
-	}
-
 	// Hidden is almost always specified, need to look at the actual value.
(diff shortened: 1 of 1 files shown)


FreeCap23:
Hi @emersion, could I get a review of this PR please? I've been using this fork since I made it and I haven't encountered any issues, though I recognize simply removing this check probably breaks something else that I don't know about. I'm open to modifying it further.

Thanks.

emersion:
I don't think this makes sense: https://github.com/emersion/mako/issues/413#issuecomment-3622833301

Issue #614, opened 2025-12-02
[contrib] provide an OpenRC (user) service


fkobi:
OpenRC provides user services since 0.60.
It would be nice to have mako available there.

PR #615, opened 2025-12-02
Add OpenRC user init scrpt

fkobi:
OpenRC supports user services since 0.60.
This commit adds a simple service script that mimics what the systemd service is doing.

Wrote & tested on Gentoo, inspired by gui-apps/foot's.

Please note that checking $WAYLAND_SESSION requires `rc_env_allow=WAYLAND_DISPLAY` to be set in user's `/rc.conf`

Closes: https://github.com/emersion/mako/issues/614


emersion (approved):
Thanks!

PR #639, opened 2026-04-08
Add timeout-nodismiss config option and unread tracking

HollowPhobos:
# Problem
When notifications expire via timeout, they move to history indistinguishably from notifications the user explicitly dismissed. There's no way for external tools (waybar modules, scripts) to know whether the user actually saw and interacted with a notification.
This problem exists independently from adding `mode=away` from outside of mako.

#  Solution
Add a `timeout-nodismiss` boolean style option. When enabled, notifications that expire via timeout are moved to history with an unread flag. Explicitly dismissed notifications do not get this flag.
- New config key: `timeout-nodismiss=0|1` (per-style, criteria-compatible)
- New unread field on notifications, exposed conditionally in `makoctl history -j output` (field is omitted when false — zero output change for users who don't enable the feature)
- New `makoctl mark-read [-n id] [-a|--all]` command to clear the unread flag
- New MarkRead D-Bus method on fr.emersion.Mako

# Usage
In config set a timeout and enable timeout-nodismiss
```
default-timeout=3000
timeout-nodismiss=1
```

## View unread notifications
`makoctl history -j | jq '[.[] | select(.unread == true)]'`

## Mark all as read
`makoctl mark-read -a`

## Mark one as read
`makoctl mark-read -n 42`

# Design notes
- Minimal impact: the only behavioral change is one if in close_notification() that sets `unread=true` when reason is `EXPIRED` and `timeout-nodismiss` is enabled
- The unread field is metadata only — it does not affect notification lifecycle, timers, history limits, or rendering
- makoctl restore clears the unread flag (user is actively engaging with the notification)
- Completions updated for bash, fish, and zsh
- Man pages updated for both `mako(5)` and `makoctl(1)`

### DIFF
--- a/config.c
+++ b/config.c
@@ -110,6 +110,7 @@ void init_default_style(struct mako_style *style) {
 	style->actions = true;
 	style->default_timeout = 0;
 	style->ignore_timeout = false;
+	style->timeout_nodismiss = false;
 
 	style->colors.background = 0x285577FF;
 	style->colors.text = 0xFFFFFFFF;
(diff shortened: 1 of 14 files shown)


HollowPhobos:
closed for being ignored

PR #642, opened 2026-04-24
Migrate to keulim from gdk-pixbuf

emersion:
gdk-pixbuf is being sunset. keulim is a simple replacement: https://gitlab.freedesktop.org/emersion/keulim


subprojects/keulim.wrap
```diff
@@ -0,0 +1,3 @@
+[wrap-git]
+url = https://gitlab.freedesktop.org/emersion/keulim.git
+revision = HEAD
```
vilhalmer, line 3: Need to update the README to replace gdk-pixbuf in the dependency list. Not sure if we also want to add `meson subprojects download` to the beginning of the manual build steps to let it fetch the repo, or if there's another mechanism you prefer to recommend? Should at least note that icons will be automatically disabled during `setup` if the user doesn't take some step to grab keulim first.

I'm not really sure why `meson subprojects update` considers keulim "not used" before the first build even if `-Dicons=enabled` is passed during setup/configure. I haven't worked with subprojects before.

vilhalmer (approved):
Code lgtm. I tested with several random large PNG files I had laying around, as well as every unique format I had in `/usr/share/icons`. Everything worked except grayscale and one icon that crashed. Discussed in IRC, both were issues in keulim rather than mako.

subprojects/keulim.wrap
emersion: > Need to update the README to replace gdk-pixbuf in the dependency list.

Indeed, done!

> Not sure if we also want to add meson subprojects download to the beginning of the manual build steps to let it fetch the repo, or if there's another mechanism you prefer to recommend?

I would recommend installing from distro packages. Distros will start shipping the library on the next mako/swaybg/etc release.

> I'm not really sure why meson subprojects update considers keulim "not used" before the first build even if -Dicons=enabled is passed during setup/configure. I haven't worked with subprojects before.

`wrap_mode=nodownload` is set as a default in `meson.build` to prevent Meson from automagically downloading sources without the user explicitly opting in.

PR #644, opened 2026-05-01
Fix deprecated syntax in README commands

saufrecht:
This eliminates the message "WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated." when running meson build.


emersion (approved):
Thanks!

PR #650, opened 2026-06-27
Show actions as buttons under text

layus:
Here is another feature I have been using for a few days and found quite useful:

Each requested action is now drawn as a clickable button beneath the notification body: a left click on a button invokes that action and a right click anywhere dismisses, per the existing default bindings. Buttons are suppressed when on-button-left is bound to invoke-default-action, which keeps the old behaviour of clicking the body to fire the default action; accordingly the default left binding is changed to none. Documented in mako.5.scd under the bindings and actions options.

image


layus:
Again, AI generated, an probably not qualitative enough to merge as-is.

PR #651, opened 2026-06-29
Fix bottom margin regression

### LINKED ISSUE #647: Margin does not apply to bottom edge
The `margin` setting does not create a margin at the bottom of notifications at the bottom edge of the screen, when using a `bottom-left/center/right` anchor.
Left and right margin are correctly applied to them.
The `outer-margin` is still correctly applied to them.

mako version: 1.11.0
Compositor: sway 1.12

dstapl:
Reapplies 4eb73fc which applies the bottom margin to the last notification.

Closes: https://github.com/emersion/mako/issues/647

### DIFF
--- a/render.c
+++ b/render.c
@@ -480,5 +480,5 @@ void render(struct mako_surface *surface, struct pool_buffer *buffer, int scale,
 	}
 
 	*rendered_width = max_width;
-	*rendered_height = total_height;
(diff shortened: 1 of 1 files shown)


emersion (approved):
It seems like this got lost in 5e1da120ecf1595d7d216eb24ba810e92cbd7718? I don't see a reason why it got dropped there.

LGTM!

